PR #92, opened 2021-10-01
Notify accounts on email changed or account deleted

cj1993:
## What?

-Integrated publishing API
-Fixed all integrations tests
-Fixed bug introduced by ip changes
-Add env var for publishing API to pipeline

## Why?
Notify accounts on email changed or account deleted


.env.sample
```diff
@@ -4,4 +4,6 @@ SESSION_EXPIRY=
 SESSION_SECRET=
 OIDC_CLIENT_ID=
 AM_API_BASE_URL=
-OIDC_CLIENT_SCOPES=
\ No newline at end of file
+OIDC_CLIENT_SCOPES=
+GOV_PUBLISHING_API_BASE_URL_INT=
```
sanisc, line 8: you only need on value here as it is changed in the pipeline. I don't see this added to the pipeline either?

JHjava (approved):

JHjava (approved):

PR #162, opened 2021-10-20
GOVSI-888: Get service domain from domain terraform state

chrisclayson:
## What?

- Remove hard-coded DNS domain references and replace with terraform_remote_state pointing at the domain terraform

## Why?

We are swapping domains over and this saves us having to manually replace values

### DIFF
--- a/ci/terraform/user-provided-services.tf
+++ b/ci/terraform/user-provided-services.tf
@@ -18,7 +18,7 @@ resource "cloudfoundry_user_provided_service" "idp" {
   credentials_json = jsonencode({
     client_id   = random_string.account_management_client_id.result
     client_name = "${var.environment}-account-managment"
-    idp_url     = "https://api.${var.cf_domain}"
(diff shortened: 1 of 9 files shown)


ci/terraform/user-provided-services.tf
```diff
@@ -18,7 +18,7 @@ resource "cloudfoundry_user_provided_service" "idp" {
   credentials_json = jsonencode({
     client_id   = random_string.account_management_client_id.result
     client_name = "${var.environment}-account-managment"
-    idp_url     = "https://api.${var.cf_domain}"
+    idp_url     = "https://api.${local.service_domain}"
```
sanisc, line 21: is this right? wasn't it going to be oidc.

sanisc (approved):

ci/terraform/user-provided-services.tf
chrisclayson: Well spotted!

PR #216, opened 2021-11-11
Fixed sonar code smells

sanisc:
Fix sonar code smells.

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -7,3 +7,5 @@ sonar.tests=test/
 
 sonar.sourceEncoding=UTF-8
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
+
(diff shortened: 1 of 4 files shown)


sonar-project.properties
```diff
@@ -7,3 +7,5 @@ sonar.tests=test/
 
 sonar.sourceEncoding=UTF-8
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
+
+sonar.exclusions=src/assets/javascript/**
```
dbes-gds, line 11: should this be excluded given that we wrote the js?

sanisc: i have made it more restrictive as all is something that is not owned by this team

dbes-gds (approved):

PR #245, opened 2021-11-25
BAU: read environment variables in docker-compose

dbes-gds:
## What?

Read environment variables in docker-compose.

When starting-up the values can be read from a local .env file that is not in the repo.  Use .env.sample as a starting point.

## Why?

Makes docker-compose.yml a static file, configuration in a separate .env file.
Ability to apply different configurations locally when starting up.


alan-gds (approved):
Nice

PR #249, opened 2021-11-29
GOVSI-1141: log session id on am frontend and pass to apis

dbes-gds:
## What?

- Log session id on am frontend and pass to apis.
- Tweak am frontend response logging to remove unavailable fields.
- Add 'from' field to request logging to bring into line with frontend.
- Change local node debug port to stop clash with frontend.

## Why?

- Ability to view a single session across both frontend and account management.
- Some fields being logged in the am frontend response were not actually available so were never being logged.

## Related PRs

Please include links to PRs in other repositories relevant to this PR.
Delete this section if not needed.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,7 +22,7 @@ services:
     build: .
     ports:
       - 6001:6001
-      - 9230:9230
+      - 9240:9230
     volumes:
       - ./:/app
     environment:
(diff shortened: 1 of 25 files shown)


JHjava:
Is it worth adding the session-id to any of the existing unit/integration tests to validate that it is being sent?

JHjava (approved):

PR #264, opened 2021-12-15
Implement show password

sanisc:
For consistency add the show password feature on all password inputs.

### DIFF
--- a/src/components/common/layout/base.njk
+++ b/src/components/common/layout/base.njk
@@ -112,9 +112,8 @@
     }) }}
 {% endblock %}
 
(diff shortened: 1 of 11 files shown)


src/components/common/layout/base.njk
```diff
@@ -113,6 +113,7 @@
 {% endblock %}
 
 {% block bodyEnd %}
+    {% block scripts %}{% endblock %}
```
dbes-gds, line 116: should the scripts be in the block?

sanisc: yes its at the end of the body tag

PR #292, opened 2022-01-12
AUTH-1273: create separate docker image for fargate

dbes-gds:
## What?

Create separate docker image for fargate.

- Rename existing local image and update docker-compose.yml to use it.
- Add a new dockerfile to run the application in fargate.  This image needs all files locally as it cannot rely on a mounted volume.
- Use 'cp' instead of 'copyfiles' in one place in package.json as copyfiles was hanging when building and/or running the docker image.

## Why?

A separate image is needed to run in fargate as we are currently using a CF node buildpack rather than a docker image.  The existing local image could not be run in fargate as-is.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,20 @@ FROM node:16.8.0-alpine
 ENV NODE_ENV "development"
 ENV PORT 6001
 
-VOLUME ["/app"]
 WORKDIR /app
+
+COPY package.json ./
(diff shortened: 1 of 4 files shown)


JHjava:
Looks good. Minor suggestion - It would be useful to add some of the info from the PR into the commit

JHjava (approved):

PR #477, opened 2022-05-25
Remove cookies and feedback link

alex9smith:
## What?

Remove the cookies and feedback link from the manage your account page.

## Why?

We are removing the stored cookie and feedback consent attributes from
GOV.UK's Account API and the management page in the parts of the account
home area that are served from GOV.UK's Frontend app [[1]].

Before we can remove that page, we need to remove the link to it here.

This is the only link to this page, so we can also remove the config,
content and environment variables for it.

## Related PRs

https://github.com/alphagov/frontend/pull/3225

[1]: https://github.com/alphagov/frontend/pull/3225

---

**NOTE:** I can't run this app locally or run the integration tests (unit tests pass fine) so the reviewer will probably need to spin it up and check the page still renders ok.

---
[Trello](https://trello.com/c/ZmuMTQy5/1370-remove-cookies-and-feedback)


dbes-gds:
There are a couple more places where it would be good to tidy up the pipeline:

COOKIES_AND_FEEDBACK_URL: ((production-cookies-and-feedback-url))
COOKIES_AND_FEEDBACK_URL: ((staging-cookies-and-feedback-url))

dbes-gds (approved):

PR #530, opened 2022-07-12
AUT-198: Add cookie which tracks manual logout

DTaylorGDS:
## What?

- Add cookie called `lo` (short for logged_out -> used abbreviation for consistency with other cookie names) to logout controller
- Read the status of that cookie in requires auth middleware
- Change the logic in that middleware so that instead of always either passing to `next()` or routing to session expired, we can now also direct to a new signed out page, based on the ability the `lo` cookie gives to distinguish between the two situations in which there is not a valid session
- PR includes creation of that new `/signed-out` page - which is very similar to the sign-in journey's `/signed-out` page, but differs lightly (see below) hence the need for a new route/view rather than reusing the same URL as initially implemented:

**Existing signed-out page for frontend:**
image

**New proposed signed-out page specific to account-management.** 
Primary difference is wording around going 'back' and the fact that the sign-in link routes via the account management 'START' url:
image

## Why?

- This allows differentiation of active logout (redirecting user to generic sign-in screen) vs session timeout/expiry (redirecting user to a screen informing them of timeout).

### DIFF
--- a/src/components/logout/logout-controller.ts
+++ b/src/components/logout/logout-controller.ts
@@ -3,5 +3,6 @@ import { Request, Response } from "express";
 export async function logoutGet(req: Request, res: Response): Promise<void> {
   const idToken = req.session.user.tokens.idToken;
   req.session.destroy();
+  res.cookie('lo', 'true');
   res.redirect(req.oidc.endSessionUrl({ id_token_hint: idToken }));
 }
--- a/src/components/logout/logout-controller.ts
+++ b/src/components/logout/logout-controller.ts
@@ -3,6 +3,6 @@ import { Request, Response } from "express";
 export async function logoutGet(req: Request, res: Response): Promise<void> {
   const idToken = req.session.user.tokens.idToken;
   req.session.destroy();
-  res.cookie('lo', 'true');
+  res.cookie("lo", "true");
   res.redirect(req.oidc.endSessionUrl({ id_token_hint: idToken }));
 }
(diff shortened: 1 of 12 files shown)


src/components/signed-out/signed-out-routes.ts
```diff
@@ -0,0 +1,10 @@
+import { PATH_DATA } from "../../app.constants";
+
+import * as express from "express";
```
JHjava, line 3: possible to avoid the wildcard import?

src/components/logout/logout-controller.ts
```diff
@@ -3,5 +3,6 @@ import { Request, Response } from "express";
 export async function logoutGet(req: Request, res: Response): Promise<void> {
   const idToken = req.session.user.tokens.idToken;
   req.session.destroy();
+  res.cookie("lo", "true");
```
JHjava, line 6: is the value for this cookie reset on login?

DTaylorGDS: Directly, no, but in practice yes. It's reset the first time the auth middleware is called with a valid, authenticated user session - which would be after a successful login and the first time the user tries to access an auth-protected page like the main account management page. Does that work for you?: https://github.com/alphagov/di-authentication-account-management/blob/f3136dbbea03b9618e24a9344496c8bed66f05d0/src/middleware/requires-auth-middleware.ts#L17

src/components/signed-out/signed-out-routes.ts
DTaylorGDS: Sure - changed to named import of Router, as only this is used

src/components/logout/logout-controller.ts
JHjava: Ah ok. Yep. That makes sense.

JHjava (approved):

dbes-gds (approved):

PR #534, opened 2022-07-14
AUT-557: change footer cookie href

DTaylorGDS:
## What?

- Change footer cookie href from generic gov.uk cookies to signin custom cookies page
- Change cookie banner 'view cookies' href from generic gov.uk cookies to signin custom cookies page

## Why?

- So that user has correct set of cookie information

### DIFF
--- a/src/components/common/layout/base.njk
+++ b/src/components/common/layout/base.njk
@@ -9,14 +9,10 @@
 
     {# For Internet Explorer 8, you need to compile specific stylesheet #}
     {# see https://frontend.design-system.service.gov.uk/supporting-ie8/#support-internet-explorer-8 #}
-    <!--[if IE 8]>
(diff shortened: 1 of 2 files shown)


src/components/common/layout/base.njk
```diff
@@ -92,7 +87,7 @@
                     text: 'general.footer.accessibilityStatement.linkText' | translate
                 },
                 {
-                    href: "https://www.gov.uk/help/cookies",
+                    href: "/cookies",
```
dbes-gds, line 90: cookies page is on the frontend so I think it's:

href: authFrontEndUrl + "/cookies"

DTaylorGDS: Yes, you're right of course. Thanks

dbes-gds (approved):

PR #542, opened 2022-07-21
GUA-226: Update content on delete account page

danacotoran:
~**This can be reviewed but please do not merge yet, we will deploy this once we hear from LITE regarding the exact date/time they are going live (due to onboard 25 July, but there’s a chance there might be delays).**~
LITE have gone live with GOV.UK Sign In so we can go ahead. 🚀 

## What?

Update content of 'Are you sure you want to delete your account' screen ahead of LITE service onboarding.

Design is here: https://www.figma.com/file/tepd26ZAc8QhiLJlxYN1wR/Data-management-and-Navigation?node-id=1522%3A3976

### Before 
Screenshot 2022-07-21 at 13 55 22

### After 
Screenshot 2022-07-22 at 09 58 58

## Why?
The purpose of these updates is so that the page more clearly explains the consequences of deleting your account.

### DIFF
--- a/src/components/delete-account/delete-account-controller.ts
+++ b/src/components/delete-account/delete-account-controller.ts
@@ -6,10 +6,12 @@ import { PATH_DATA } from "../../app.constants";
 import { getNextState } from "../../utils/state-machine";
 import { GovUkPublishingServiceInterface } from "../common/gov-uk-publishing/types";
 import { govUkPublishingService } from "../common/gov-uk-publishing/gov-uk-publishing-service";
-import { getBaseUrl } from "../../config";
+import { getBaseUrl, getManageGovukEmailsUrl } from "../../config";
 
 export function deleteAccountGet(req: Request, res: Response): void {
-  res.render("delete-account/index.njk");
+  res.render("delete-account/index.njk", {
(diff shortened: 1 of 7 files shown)


src/app.constants.ts
```diff
@@ -3,7 +3,9 @@ import { UserJourney } from "./utils/state-machine";
 export const PATH_DATA: {
   [key: string]: { url: string; event?: string; type?: UserJourney };
 } = {
+  GOVUK_HOMEPAGE: { url: "https://www.gov.uk" },
```
alex9smith, line 6: Setting a variable is the right thing to do, but I think this is the wrong place to put it. `PATH_DATA` is all the internal paths and is used by the state machine to work out where someone's allowed to go next. 

I think it'd be better to put a `getManageGovUkEmailsUrl()` to [config.ts](https://github.com/alphagov/di-authentication-account-management/blob/main/src/config.ts) and bring that into your controllers.

danacotoran: Good point @alex9smith, thanks for explaining the difference

alex9smith (approved):
👍🏻

PR #555, opened 2022-08-05
BAU: Publish logs to multiple CSLS endpoints

mrwilson:
Migrating to new CSLS endpoint

### DIFF
--- a/ci/tasks/deploy-account-management.yml
+++ b/ci/tasks/deploy-account-management.yml
@@ -56,6 +56,7 @@ run:
         -var "gov_accounts_publishing_api_url=${PUBLISHING_API_URL}" \
         -var "gov_account_publishing_api_token=${PUBLISHING_API_URL_TOKEN}" \
         -var "logging_endpoint_arn=${LOGGING_ENDPOINT_ARN}" \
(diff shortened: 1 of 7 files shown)


ci/tasks/deploy-account-management.yml
```diff
@@ -56,6 +56,7 @@ run:
         -var "gov_accounts_publishing_api_url=${PUBLISHING_API_URL}" \
         -var "gov_account_publishing_api_token=${PUBLISHING_API_URL_TOKEN}" \
         -var "logging_endpoint_arn=${LOGGING_ENDPOINT_ARN}" \
+        -var 'logging_endpoint_arns=["arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prod"]' \
```
dbes-gds, line 59: is this needed? it will overwrite what's in tfvars

mrwilson: I thought tf vars, because it follows this, will override the default on this line? This is what we do in di-authentication-api 

https://github.com/alphagov/di-authentication-api/blob/main/ci/tasks/deploy-oidc-api.yml#L60

dbes-gds: right you are then...

dbes-gds (approved):

PR #569, opened 2022-08-18
AUT-554: Make govukInput elements numeric for OTP code

DTaylorGDS:
## What?

- Add fields to all govukInput elements which take an OTP code
- Make input mode numeric and pattern match for "[0-9]*"
- This is instead of <input type="number"> which can be increased accidentally with keyboard arrows or scroll wheel

## Why?

- So that users are less likely to submit an incorrect OTP code simply by having knocked their mouse or keyboard and therefore changing their input without noticing

## Related PRs

Same thing for frontend: https://github.com/alphagov/di-authentication-frontend/pull/723

### DIFF
--- a/src/components/check-your-email/index.njk
+++ b/src/components/check-your-email/index.njk
@@ -9,23 +9,23 @@
 
 {% block pageContent %}
 
-{% include "common/errors/errorSummary.njk" %}
+    {% include "common/errors/errorSummary.njk" %}
 
-<h1 class="govuk-heading-l govuk-!-margin-top-0 govuk-!-margin-bottom-3">{{'pages.checkYourEmail.header' | translate}}</h1>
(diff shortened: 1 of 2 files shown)


gtvj (approved):

src/components/check-your-email/index.njk
```diff
@@ -9,52 +9,52 @@
...
+        {{ govukInput({
     label: {
         text: 'pages.checkYourEmail.code.label' | translate
     },
     classes:"govuk-input--width-5",
     id: "code",
     name: "code",
     value: code,
+    inputmode: "numeric",
+    pattern: "[0-9]*",
```
dbes-gds, line 37: does it help if this is limited to six chars? or would it do weird stuff?

gtvj: I'm not entirely sure but I can ask the Design System team. The pattern used here seemed to fit with what's presented in the Design System [example where there is a specified character range](https://design-system.service.gov.uk/components/text-input/#numbers).

PR #585, opened 2022-08-30
AUT-200: Account Management Welsh translations

dbes-gds:
## What?

Account Management Welsh translations.

- Upload the translations file as received.
- Add in translations for 'securityCodeInvalid' which were added to the English before the file was sent for translation.
- Remove i18n 'header' configuration to align with frontend.
- Switch on in Staging.
- Do not load Welsh in any environment where it is not switched on.  This prevents anyone from changing the language.

## Why?

A Welsh language version of the application is required.
Enable first view testing of the translation.

In Account Management i18n config included 'header' which was different from the Frontend config.  This meant that browser language settings would override the cookie, switching the language back when it had been changed.  This has been removed so that language choice is driven by query parameter then cookie as per Frontend.

## Related PRs

https://github.com/alphagov/di-authentication-frontend/pull/737
https://github.com/alphagov/di-authentication-api/pull/2292


docker-compose.yml
```diff
@@ -45,4 +45,5 @@ services:
       - ANALYTICS_COOKIE_DOMAIN=${ANALYTICS_COOKIE_DOMAIN}
       - GTM_ID=${GTM_ID}
       - SUPPORT_INTERNATIONAL_NUMBERS=${SUPPORT_INTERNATIONAL_NUMBERS}
+      - SUPPORT_LANGUAGE_CY=${SUPPORT_LANGUAGE_CY}
```
chrisclayson, line 48: Do we also need to add `SUPPORT_LANGUAGE_CY` to `.env.sample`?

dbes-gds: yes good call will add

chrisclayson (approved):

PR #654, opened 2022-10-27
GUA-477 ADR Read from service store

huwd:
## What?

Detail plans to read data directly from the store

## Why?

We'd not actually documented this anywhere

## Related PRs
https://github.com/alphagov/di-authentication-account-management/pull/652


docs/adr/0003-read-data-out-of-account-store-into-app.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Decision
+
+We will use the AWS SDK to directly query the data store from the express application.
+
+![Architecture diagram showing direct querying of the dynamo](./images/2022-reading-data-out-adopted.png)
+
+We picked this option as it is extremely simple, requiring one dependency added to the app, and no extra AWS components.
+
+As a downside it depends on this application and the dynamo being within the same AWS Account.
```
alex9smith, line 16: That's not correct - we can set up cross account access to query a DynamoDB table in another account. Take a look at this AWS guide for [cross account access for Dynamo](https://docs.aws.amazon.com/prescriptive-guidance/latest/patterns/configure-cross-account-access-to-amazon-dynamodb.html).

alex9smith: Which is good! It makes the argument between the options purely a decision between complexity and coupling, and it seems pretty reasonable to choose the low complexity option to speed up implementation at the cost of coupling two services we don't expect to change much

huwd: thanks for that, didn't know there was cross account dynamo access, very interesting!

huwd:
Taking this off draft, let's review this now.
if we're happy I can always extend later if there are comments from Paul?

alex9smith (approved):

PR #662, opened 2022-11-02
GUA 473: Add new header to the account

danacotoran:
Add new header design including the new brand name as well as links to the services page (which does not exist yet).
The new design is behind a flag for now seeing as we might want to switch everything on (the account area redesign and header redesign) all at the same time.

[Latest design on Figma](https://www.figma.com/file/Q6n44LmHFsCOGzZ44QTW4U/Account-home?node-id=2281%3A14816)

Not ready for review yet as we need to be able to deploy it to a testing environment as well as Welsh translations for the new content.


huwd:
Once this is in:
https://github.com/alphagov/di-authentication-account-management/pull/607

I'll add a feature flag to this, so we can merge?

huwd:
@danacotoran I think this is ready for a review from @alex9smith and @divyaparameswaran.

I think let's get to a point where we can see this working on Dev before we actually merge.
But we should have confidence in the feature flags to prevent us showing anything to the user.

So we could do a follow up PR if there is still outstanding Welsh Translation needed?

danacotoran:
@huwd should we hold off on the code review until it's been approved by the UCD team? As there could be changes resulting from the design review.

alex9smith (approved):
LGTM. I think I'd prefer to merge this in behind the feature flag and make any follow up changes in a separate PR

huwd (approved):

huwd (dismissed):

huwd (approved):

PR #673, opened 2022-11-08
AUT-879 - Delete all sessions when account is deleted

JHjava:
## What?

- When a user has deleted their account we should remove all sessions assoicated with that user.

## Why?

- Otherwise a user can still have an open session within account management and then when the user goes to perform an action the system will throw an error.


huwd:
could we hold off merging this until our frontender @danacotoran has had a chance to review.
We're pretty imminently about to take ownership of this repo, I want to make sure we're happy to maintain this + test coverage etc.

At a glance, looks good though :)

huwd (changes requested):
Abusing the request changes feature to ensure @danacotoran gets a chance to 👀 

Give me a shout on slack if this is a problem!

huwd (approved):

PR #683, opened 2022-11-15
BAU: Remove seed.yaml

huwd:
Seed.yaml is in the .gitignore but was tracked anyway.
We should replace it with a sample, some instructions on how to get one, and remove the tracked file.


seed.yaml.sample
```diff
@@ -0,0 +1,10 @@
+Keys:
+  Asymmetric:
+    Rsa:
+      - Metadata:
+          KeyId: "REPLACE_ME_KEY_ID_HERE"
+          KeyUsage: SIGN_VERIFY
+          Description: RSA key with 2048 bits
+        PrivateKeyPem: |
+          -----BEGIN PRIVATE KEY-----
```
huwd, line 9: Urgh GitGuardian is emailing me to say I've _leaked a private key_ because of this commit.

Guess we know how their Regex works now.

seed.yaml.sample
```diff
@@ -0,0 +1,10 @@
+Keys:
+  Asymmetric:
+    Rsa:
+      - Metadata:
+          KeyId: "REPLACE_ME_KEY_ID_HERE"
```
alex9smith, line 5: The key ID will need to be the same as this value: https://github.com/alphagov/di-authentication-account-management/blob/main/src/config/aws.ts#L10

README.md
```diff
@@ -8,40 +8,44 @@
...
 
 ### Setup the private key
 
-Copy the private key generated in step one and put it in `seed.yaml` with the correct identation.
+Create a copy of the seed.yaml.sample, rename it seed.yaml and fill in the value for the private key, using the key generated above or shared by a colleague.
 
 ### Update the WAF
 
-The WAF will block attempts made by the local application to use the build api unless some rules are switched off.  This can be done temporarily in the build environment during testing.  When a deployment happens the rules will be reset, but it's a good idea to keep the rules the same in all environments as much as possible, so these rules have not been switched off permanently in build.
+The WAF will block attempts made by the local application to use the build api unless some rules are switched off. This can be done temporarily in the build environment during testing. When a deployment happens the rules will be reset, but it's a good idea to keep the rules the same in all environments as much as possible, so these rules have not been switched off permanently in build.
```
alex9smith, line 48: I'm still a bit nervous about this, but it seems to work so let's leave it for now and come back if we have any problems.

huwd: Ah, I didn't set out to change this 🤔 I think this is my text linter getting rid of whitespace...

huwd: Yeah so this came in as content here:
https://github.com/alphagov/di-authentication-account-management/commit/93a19ac35a5059f29a73e0d7557aa216d583f088

I'm guessing your nerves are about the content and not the whitespace linting @alex9smith

seed.yaml.sample
huwd: Ok cool, I'll hard code that in then

huwd: Or... actually humm...

Having this hard coded in two places kind of sucks.
It might be better if [this line](https://github.com/alphagov/di-authentication-account-management/blob/main/src/config/aws.ts#L10) loaded the YAML and took the KeyID from this file, no?

huwd: point about not hard coding it in seeds.yaml is that someone could register a new client_id with auth build?

README.md
alex9smith: Ah yeah, makes sense! Yup I'm concerned about the content but like I said it seems to work fine

seed.yaml.sample
alex9smith: Setting the key ID is just to make sure the app is using the correct key when it calls out to localstack - that ID won't need to change if someone registers a new client to use locally. 

I see what you mean about having to set it in two places but we shouldn't ever have to change this unless we completely change how we run this app locally. I think it's fine as is in this case

huwd: Ok gotcha, that's a lot easier then

huwd: Done
https://github.com/alphagov/di-authentication-account-management/pull/683/files#diff-bbbe0411a120b1a494f185ef61d6256dddd80c9176ae4485a22f62a4de76bc00R5

PR #689, opened 2022-11-16
BAU: Tag images for the dev environment with the commit SHA

alex9smith:
## Proposed changes
Previously this action would tag the Docker image as 'latest'. ECR handles this ok by saving the newly uploaded image as 'latest' and untagging the image that was previously 'latest'.

However when the CloudFormation template is applied, the running image in ECS isn't updated. I think this is for one of two reasons:

1. CloudFormation sees that ECS is already running an image tagged 'latest' and decides not to update it
2. ECS caches images and when the new task definition is started it uses the already cached 'latest' image, which isn't what we want.

Both of these can be prevented by tagging each image with a unique ID.

### What changed

This PR changes the action to tag the image with the commit SHA in the same way as the action which uploads an image for the secure pipeline.

This means we no longer need the condition or the hard coded image value for the dev environment.

### Why did it change

This action didn't deploy the new image to ECS. 

### Other considerations

### DIFF
--- a/.github/workflows/cd-only.yml
+++ b/.github/workflows/cd-only.yml
@@ -60,10 +60,14 @@ jobs:
         id: publish-template
         env:
           ARTIFACT_BUCKET_NAME: ${{ secrets.DEV_ARTIFACT_SOURCE_BUCKET_NAME }}
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: ${{ secrets.DEV_ECR_REPOSITORY }}
         run: |
           echo "Running sam build on template file"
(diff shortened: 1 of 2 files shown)


alex9smith:
Some proof that this works

Successful action run: https://github.com/alphagov/di-authentication-account-management/actions/runs/3480708379

Tagged image in ECR:
image

Task definition with new image: 
image

(blanked out some IDs just in case)

alex9smith:
I've put this back in draft - after some discussion with @TaherKapasi we don't think this is what's causing the problem and removing the hard-coded config for dev does bring some additional problems.

Let's wait on doing anything with this until we've tried out something else

PR #690, opened 2022-11-16
GUA-470 Trigger delete

huwd:
WIP:

I _think_ this might work.
I'm not really happy with the test coverage, but was trying to go with what's already there.

I presume the focus here is on integration testing, so I think we'll need an integration test to check we get an AWS event fired off to SNS...

But might try manually giving it a go on dev first.


huwd:
I couldn't think up a better way of testing the feature flag. I know this adds another massive section of tests... But I have a feeling we should get something that works now, and then perhaps review the approach to this app over all?

This service pattern is adding a lot of boilerplate code, and I'm uncertain if we're getting value from it right now?

divyaparameswaran (approved):

PR #699, opened 2022-11-18
Revert "AUT-895: Create a redirection lambda function"

chrisclayson:
Reverts alphagov/di-authentication-account-management#696


JY0280 (approved):

huwd:
@chrisclayson We've switched all our infra as code here to use Cloud Formation, following Dev Platform's lead.
My assumption was once we'd switched over, we'd eventually clear out the CI folder and all the old Auth Terraform.

Is this lambda going to be something you'll want to maintain after the switchover date?

If so, should we consider a different repo for this?
Or should we look to move this into cloud-formation for ongoing maintenance?

chrisclayson:
> @chrisclayson We've switched all our infra as code here to use Cloud Formation, following Dev Platform's lead. My assumption was once we'd switched over, we'd eventually clear out the CI folder and all the old Auth Terraform.
> 
> Is this lambda going to be something you'll want to maintain after the switchover date?
> 
> If so, should we consider a different repo for this? Or should we look to move this into cloud-formation for ongoing maintenance?

@huwd ... this is very much just a short-term solution to accelerate the switch over, my hope is by the time we clear out the CI folder, a more sustainable way of doing this will be in place.

PR #706, opened 2022-11-21
ATB-125: Enable Branch deploys

huwd:
Enable branch deploys

### DIFF
--- a/.github/workflows/build-test-application.yml
+++ b/.github/workflows/build-test-application.yml
@@ -17,6 +17,11 @@ on:
     branches-ignore:
       - main
   workflow_call:
+    inputs:
+      gitRef:
+        required: false
+        type: string
+        default: ${{ github.ref }}
 
 jobs:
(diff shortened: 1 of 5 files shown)


alex9smith:
Could you add some more detail to the PR / commit description about what this is doing and why it's needed? I suspect that'll be useful when we come to do something similar for the backend

alex9smith:
I've tested this with the branch to enable the new header and it works 🎉 

Could you also update the instructions in the README to reflect the new process. We now need to run the action off `main` in the dropdown and put our branch name in the input box.

alex9smith:
Ah, my bad - we already have some instructions in the README ([see here](https://github.com/alphagov/di-authentication-account-management#deploying-to-the-development-aws-account)). I should have linked to those in my previous comment. 

Could you consolidate the two sets? I don't mind which one you go with

alex9smith (approved):

PR #707, opened 2022-11-21
GUA-503: Show the user's new email when sending a new code

alex9smith:
## Proposed changes

Show the user's new email address when sending a new verification code.

Note - this doesn't change any behaviour about which email we send the code to. That's already working as intended - it already correctly goes to their new email address.

This change only switches the email displayed to the user to the right one.

### Why did it change

The only way a user can get a verification code sent to their email address by this app is while they're going through the journey to change their email address. The other credential and delete journeys require them to re-enter their password, but we never send a code to their email.

That means the only way a user can end up resending an email code is if they're trying to send a code to their new email address while they're in the process of switching. We should display their new email address on this page, but we're showing the old one. This is confusing, because the verification code gets sent to their new email address.

### Issue tracking
https://govukverify.atlassian.net/browse/GUA-503

### Other considerations

### DIFF
--- a/src/components/resend-email-code/resend-email-code-controller.ts
+++ b/src/components/resend-email-code/resend-email-code-controller.ts
@@ -14,7 +14,7 @@ const TEMPLATE_NAME = "resend-email-code/index.njk";
 
 export function resendEmailCodeGet(req: Request, res: Response): void {
   res.render(TEMPLATE_NAME, {
-    emailAddress: req.session.user.email,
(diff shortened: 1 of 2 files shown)


huwd (dismissed):
huh, good spot

huwd:
Bit nervous that doesn't break any tests...

alex9smith:
Good point, I've added a simple test for it

huwd (approved):

PR #709, opened 2022-11-21
AUT-895: Fix deployment task

chrisclayson:
## Proposed changes

Fix Terraform deployment task for redirect lambda

### What changed

- Add `redirect_lambda_zip` Terraform variable to deploy task

### Why did it change

Deployment pipeline is failing

### Issue tracking

- [AUT-895](https://govukverify.atlassian.net/browse/AUT-895)

### DIFF
--- a/ci/tasks/deploy-account-management.yml
+++ b/ci/tasks/deploy-account-management.yml
@@ -27,6 +27,7 @@ params:
 inputs:
   - name: account-management-src
   - name: account-management-image
+  - name: acct-mgmt-redirect-release
 outputs:
   - name: terraform-outputs
 run:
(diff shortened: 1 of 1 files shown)


TaherKapasi:
Hi @chrisclayson - There is an intension to eventually delete the terraform assets along with the old pipeline files from this repo now that we have a deployment through secure pipelines with SAM.  The deletion would occur after the switchover has happened and after all the assets/resources have been after the destroyed.  However, I presume the redirect feature needs to exist afterwards and am wondering if it should actually be in another repo?

JY0280 (approved):

chrisclayson:
> Hi @chrisclayson - There is an intension to eventually delete the terraform assets along with the old pipeline files from this repo now that we have a deployment through secure pipelines with SAM. The deletion would occur after the switchover has happened and after all the assets/resources have been destroyed. However, I presume the redirect feature needs to exist afterwards and am wondering if it should actually be in another repo?

Hey @TaherKapasi ... yes, this is fully understood, this was 'quick win' implementation to get the switchover done, I am working on the long term solution as separate stream.

PR #721, opened 2022-11-24
ATB- 109 Add required infrastructure for DynamoDB

MinaMahdi:
### What?
The account management application currently uses Shared-Redis as a session store via Elasticache and we have decided to move to using DynamoDB for session store instead, so have created the required infrastructure for the move to DynamoDB. 

### Why? 

To kickstart the migration from Shared- Redis to DynamoDB for the purpose of decreasing costs overall.

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1394,10 +1394,7 @@ Resources:
               - "kms:ReEncrypt*"
               - "kms:GenerateDataKey*"
               - "kms:Describe*"
-            Resource: "*"
-            Condition:
-              ArnLike:
-                aws:SourceArn: !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:/${AWS::StackName}"
+            Resource: !Sub "arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${AWS::StackName}-SessionsDynamoDB"
       Tags:
         - Key: Product
           Value: !Ref ProductTagValue
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1334,17 +1334,13 @@ Resources:
     Properties:
       AttributeDefinitions:
         AttributeDefinition:
-        -
-          AttributeName: "id"
+        - AttributeName: "id"
           AttributeType: "S"
-        -
-          AttributeName: "expires"
+        - AttributeName: "expires"
           AttributeType: "N"
-        -
-          AttributeName: "type"
+        - AttributeName: "type"
           AttributeType: "S"
-        -
-          AttributeName: "sess"
+        - AttributeName: "sess"
           AttributeType: "S"
       #          AttributeName: "subjectId"
       #          AttributeType: "S" - part of GSI
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -314,7 +314,6 @@ Resources:
         - Key: Source
           Value: !Ref SourceTagValue
 
-
   ECSClusterNameParameter:
     Type: AWS::SSM::Parameter
     Properties:
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1333,40 +1333,23 @@ Resources:
     Type: AWS::DynamoDB::Table
     Properties:
       AttributeDefinitions:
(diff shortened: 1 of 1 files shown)


TaherKapasi:
Hi @MinaMahdi, @domantas-gds & @KG700 - is the table name going to be passed to the Container via env var setting?

MinaMahdi:
> Hi @MinaMahdi, @domantas-gds & @KG700 - is the table name going to be passed to the Container via env var setting?

Hi @TaherKapasi , yes as far as I've understood that's something that is going to be passed.

deploy/template.yaml
```diff
@@ -1325,6 +1325,99 @@ Resources:
...
+          - Effect: Allow
+            Principal:
+              Service: !Sub "dynamodb.amazonaws.com"
+            Action:
+              - "kms:Encrypt*"
+              - "kms:Decrypt*"
+              - "kms:ReEncrypt*"
+              - "kms:GenerateDataKey*"
+              - "kms:Describe*"
+            Resource: "*"
```
huwd, line 1397: 🤔 could we tighten up the resource definition here?
This is for the Dynamo right? Could we use a `!Ref SessionsDynamoDB` here from the above session to narrow the scope a bit?

deploy/template.yaml
```diff
@@ -1325,6 +1325,99 @@ Resources:
...
+      EnableKeyRotation: true
+      KeyPolicy:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal:
+              AWS: !Sub "arn:aws:iam::${AWS::AccountId}:root"
+            Action:
+              - kms:*
+            Resource: "*"
```
huwd, line 1387: 🤔 could we have a bit of a chat about this one.
This seems kinda broad, I'm still a bit of a newbie to AWS permissions but this seems to be a broad KMS permission, applied to all resources. I suspect we could narrow that a bit.

Also not sure the ffect of L1383/L1384, but `root` on the account ID as a principle makes me 🤔

huwd (commented):
Looks good, a couple of questions about if we can narrow some permissions there?

deploy/template.yaml
domantas-gds: Hey @huwd, we can't do an explicit `!Ref SessionDynamoDB` as the DynamoDB references the KMS key, so a KMS key to DynamoDB reference would cause a cyclic dependency where each resource is dependent on the creation of the other. I see Mina has defined a condition which only allows resources of ArnLike `arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:/${AWS::StackName}` but we can shift the ARN to the Resource block definition, remove the ArnLike condition, and slightly update the interpolated value to be `arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/${AWS::StackName}-SessionsDynamoDB` to match the ARN of the DynamoDB.

deploy/template.yaml
domantas-gds: Hi @huwd, as per documentation [here](https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html) by default AWS KMS key policies do not provide account permissions and we have therefore been providing root user access to all of our KMS resources to allow us to manage our KMS keys. That being said you may be right that having root access is too broad and instead we should potentially provide access to our admin roles to manage the keys. This affects all of our KMS related policies, not just this resource alone, so I'll create an enhancement ticket to further investigate what the best approach is and reflect on all KMS related key policies.

domantas-gds: So if you're happy to resolve this comment in scope of this work, I'll look into it from a broader KMS policy perspective.

domantas-gds: Just had a look at other repos and seems other teams are providing the same `root` permissions. Will investigate further.

domantas-gds: Hi @huwd did some investigating on the above and we'll need to keep the KMS key policy permissions as is, please see comments in ticket https://govukverify.atlassian.net/browse/ATB-155.

TaherKapasi:
Replaced by [new PR for ATB-108](https://github.com/alphagov/di-authentication-account-management/pull/780/files)

PR #725, opened 2022-11-25
GUA-541: Enable additional metrics for CloudFront

alex9smith:
CloudFront exports some metrics to CloudWatch by default, but those don't include origin latency or the number of 4XX / 5XX responses.

Enable the real time metrics export to CloudWatch which include these metrics [[1]].

[1]: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/viewing-cloudfront-metrics.html#monitoring-console.distributions-additional

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1239,6 +1239,14 @@ Resources:
           QueryStringBehavior:
             all
 
+  CloudFrontMonitoringSubscription:
+    Type: AWS::CloudFront::MonitoringSubscription
+    Properties: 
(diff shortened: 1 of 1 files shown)


alex9smith:
🤔 This failed to deploy to dev with a permissions error. I think that's because the pipeline role doesn't have permission to create this subscription. I'll ask dev platform and put this into draft for now

alex9smith:
We've decided to go ahead with metrics from API Gateway instead for now

PR #731, opened 2022-11-30
Revert "BAU: Switch off account management redirect"

alex9smith:
Reverts alphagov/di-authentication-account-management#728

We've fixed the issue that meant we needed to rollback to the old infrastructure, so we can safely restore the redirect and start serving production traffic again.


alex9smith:
I'll hold off on merging this until we've checked with Pete / Rachel

TaherKapasi (approved):

pcornish:
Following the conversation with Taher and Huw today, reenabling this is contingent on the runbook/docs being written up on how to check if the cache is being hit in an environment.

Otherwise LGTM 👍

pcornish (approved):
LGTM

PR #754, opened 2022-12-15
BAU: fix header/nav issues

danacotoran:
Firefox doesn't like the double negative margins between the DS header and the new account nav which is leading to some bizzare behaviour when zooming in and out where the nav links are jumping to the left/right at different zoom levels on Firefox.

This implements an alternative solution which instead of hiding the blue border at the bottom of the black bar by adding negative margins to the account nav, uses a modifier class which removes the bar and negative margins at the bottom of the header.

Also makes the container for the GOVUK logo and Account product name take up 100% of the width of the containing element. It was spotted that the "Account" product name collapses below the logo on certain breakpoints. Increasing the width of the container ensures this will not happen.

### Issue tracking

[GUA-575](https://govukverify.atlassian.net/browse/GUA-575)

### DIFF
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -111,12 +111,25 @@ $govuk-new-link-styles: true;
   background-color: govuk-colour("light-grey", $legacy: "grey-3");
 }
 
+// the modifier below is in place to remove the non-optional blue border at the bottom of the Design System header component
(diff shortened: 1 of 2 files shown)


huwd (approved):
Great work, thanks!

Again can we hold this until Jan 3rd

PR #756, opened 2022-12-19
GUA-522: Update dataLayer push with language

danacotoran:
Push language and language code to the data layer as per PA request. This is so Performance Analysts have access to information on the currently selected language in GA.

I shamelessly copied the [current implementation](https://github.com/alphagov/di-authentication-frontend/blob/main/src/assets/javascript/cookies.js#L113) on `di-authentication-frontend` as we want our implementation of this to work exactly the same.

### DIFF
--- a/src/assets/javascript/cookies.js
+++ b/src/assets/javascript/cookies.js
@@ -122,6 +122,25 @@ var cookies = function (trackingId, analyticsCookieDomain) {
     initGtm();
   }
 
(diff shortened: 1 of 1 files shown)


huwd (approved):

huwd:
Looks good but @danacotoran please don't merge, can we hold until we're ready after the 3rd?

PR #790, opened 2023-01-26
[ATB-207] Replaced var keyword with let or const

KG700:
### What changed

Replaced the `var` keyword with either `let` or `const`.

### Why did it change

Using the `var` keywords creates a code smell since it is outdated syntax.


KG700:
Abandoning this PR because the frontend javascript files need to use es5 to ensure support for older browsers such as IE11.

PR #794, opened 2023-01-31
Renaming the application 

DougsCER:
### What changed
Renaming the service within documentation. 
Changing links and references to the service.
Updating the name in github actions and in sonarcloud configuration 

### Why did it change
Having 'authentication' in the repo name is no longer relevant because the service does not handle and authentication responsibilities. 

### Issue tracking
GUA-401 https://github.com/alphagov/di-account-management-frontend/tree/GUA-401

### Other considerations
- [x] Update [README](./blob/main/README.md) with any new instructions or tasks


sonar-project.properties
```diff
@@ -1,4 +1,4 @@
-sonar.projectKey=alphagov_di-authentication-account-management
+sonar.projectKey=alphagov_di-management-frontend
```
divyaparameswaran, line 1: Should we add account here ?

DougsCER: good spot

divyaparameswaran (approved):

PR #806, opened 2023-02-08
[ATB-211] - Fix 'querystring' as it is deprecated

chiarapascucci-gds:
## Proposed changes
we use the stringify method of querystring to stringify an object and form a URL to redirect the user to. This module (querystring) is deprecated, so we are substituting it with [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

### What changed
We are only using querystring in src/components/oidc-callback/call-back-controller.ts, as such this is the only file changed. A new URLSearchParams object is instantiated passing the object with the query param in it. The direct URL is then formed in the same way as before but we append the output of URLSearchParams.toString() method

### Why did it change
Because querystring module became deprecated

### Issue tracking
- [ATB-211](https://govukverify.atlassian.net/browse/ATB-211)

### Local
App deployed in local docker container. Run all tests successfully 
Screenshot 2023-02-08 at 15 51 17
Screenshot 2023-02-08 at 15 53 45

Also inserted log line to manually check that output with querystring and output with URLSearchParams were the same, while running the unit tests for the oidc-callback component

output with querystring
Screenshot 2023-02-08 at 14 08 24

output with urlsearchparams
Screenshot 2023-02-08 at 15 44 45

console log was then removed

### AWS Account
Screenshot 2023-02-08 at 15 45 17

app deployed and application manually testing by logging in and ensuring that the redirect URL (/your-services) comes up as expected

### Other considerations

[ATB-211]: https://govukverify.atlassian.net/browse/ATB-211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/oidc-callback/call-back-controller.ts
+++ b/src/components/oidc-callback/call-back-controller.ts
@@ -4,7 +4,6 @@ import { PATH_DATA, VECTORS_OF_TRUST } from "../../app.constants";
 import { ExpressRouteFunc } from "../../types";
 import { ClientAssertionServiceInterface } from "../../utils/types";
 import { clientAssertionGenerator } from "../../utils/oidc";
-import * as querystring from "querystring";
 
 const COOKIES_PREFERENCES_SET = "cookies_preferences_set";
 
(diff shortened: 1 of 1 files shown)


TaherKapasi (approved):

huwd (approved):
Super interesting, didn't know about [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams), great find

PR #816, opened 2023-02-15
GUA-517: add "How you get security codes" section

danacotoran:
# What
Add new "How you get security codes" section to the settings page.

# Why 
The ability to set up an account using an authenticator app for 2fa has existed for a while. 

This displays the user's current 2fa method on their account, as well as allowing the option to change the phone number if their current 2fa method is SMS. There isn't currently an option to change auth app since that functionality does not exist on the account yet. This is something we will have to iterate on in the future.

| Text  | App |
| Screenshot 2023-02-15 at 10 23 02  | Screenshot 2023-02-15 at 10 22 28  |

### DIFF
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -13,28 +13,21 @@ $govuk-new-link-styles: true;
   }
 }
 
-.accounts-summary-list {
+.summary-list-container {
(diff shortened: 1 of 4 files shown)


divyaparameswaran (approved):
LGTM

PR #825, opened 2023-02-22
GUA-689: tweak JS builds

danacotoran:
### What changed
Concatenate cookies script into `application.js` to minimise number of network requests and uglify them for build.

Uglify govuk-frontend `all.js` script for production as the version we currently use in prod is not optimised in any way.

Change namespacing of the show password module – no real reason except that there are too many Gov-something objects on `window`.


src/assets/javascript/showPassword.js
```diff
@@ -1,5 +1,5 @@
-window.GOVUK = window.GOVUK || {};
-window.GOVUK.Modules = window.GOVUK.Modules || {};
+window.GOVSignIn = window.GOVSignIn || {};
```
huwd, line 1: Is this the only place this gets initalised in the app?
Or are there other GOVUK Module elements we'll want to clear up later?

huwd (dismissed):

src/assets/javascript/showPassword.js
danacotoran: I think the `GOVUK.Modules` thing is legacy from the days the team used to be on GOUVK as that is the standard namespacing there. This is the only place where it's referenced

huwd (approved):

PR #827, opened 2023-02-22
BAU: Add On push branch CI

huwd:
It's going to validate and perfom app checks, container checks and cloud formation checks. The checks that require the remote env, will always use our dev instance, and be separate from the actual secure pipelines CI/CD that sends things to build.

This is important as the build env will never be happy with any request that uses a git branch other than main.

We recognise that this is a bit ugly and there's lots of chances for a general refactor, but we thought we'd start by adding duplication so that it works, and then the opportunities for DRYing things out is easily testable, and a separate small unit of work.


TaherKapasi (approved):

domantas-gds:
Shall we rename `main.yml` to something more aligned with what the trigger does e.g. `on-merge-to-main.yml`

huwd:
🤔 Interesting the [Image checks fail](https://github.com/alphagov/di-account-management-frontend/actions/runs/4245303840/jobs/7380698503) when trying to assume the [DEV_GH_ACTIONS_ROLE_ARN](https://github.com/alphagov/di-account-management-frontend/pull/827/files#diff-f435406c704d1a6ed6335de88029ef371000e94a06809fdcc2e205be1c7732fcR42)

but the Sam validation checks [don't fail](https://github.com/alphagov/di-account-management-frontend/actions/runs/4245299962/jobs/7380622383) [when doing the same](https://github.com/alphagov/di-account-management-frontend/pull/827/files#diff-7537349721ddeed3219aa204cb60864a434504df6f4d7df642aaa95f7a49931cR49)...

Why is that then 🤔

huwd:
Oh scratch that, they never get that far... the one after is cancelled.

right so @domantas-gds, does this mean we need to change something in the dev instance (trust policy?) to allow branch git references to be deployed?

domantas-gds (approved):

PR #843, opened 2023-02-28
GUA - 593 Customise delete screen

divyaparameswaran:
## Proposed changes
GUA-593 : Add personalisation to the Delete screen including the services used.

### What changed

- Adjusted the delete controller to read from dynamo db similar to your-services. 
- Personalise content on the screen based on what services user has.
- Add additional unit and integration tests.

Delete screen with email subscriptions.

Delete_GOVUK

Delete without GOVUK subscriptions

Delete_withoutGOVUK


src/utils/yourServices.ts
```diff
@@ -67,3 +67,19 @@ export const formatService = (service: Service): Service => {
     last_accessed_readable_format: readable_format_date,
   };
 };
+
+export const containsGovUkPublishingService =  (serviceList: Service[]): boolean => {
+
+  // console.log(Date.now() + " DOUGAL - within govuk service calculation")
```
huwd: Interesting, what's the thinking behind leaving this commented out?

huwd: Just seen this goes in a later commit. That's fine!

huwd (approved):
Great work!
Would be good to stick a screenshot on the PR so we can see the changes?

Commits look great and are both easy to review now and good to go into the historic log. Thanks very much for the rebase there.

Reminds me we should sort out using both yarn and npm, we've got both a `yarn.lock` and `package-lock.json` change here... 🤔 not one for this PR but we should put a ticket on the backlog to choose a side!

src/locales/cy/translation.json
```diff
@@ -196,14 +196,17 @@
             "link_href": "https://exporter.lite.private-beta.service.trade.gov.uk/"
           },
           "RqFZ83csmS4Mi4Y7s7ohD9-ekwU": {
+            "header": "Gwneud cais am wiriad DBS sylfaenol",
```
danacotoran, line 199: What is the new `header` field used for?

src/locales/en/translation.json
```diff
@@ -316,18 +330,22 @@
             "link_href": "https://exporter.lite.private-beta.service.trade.gov.uk/"
           },
           "dbs": {
+             "header": "Request a basic DBS check",
```
danacotoran, line 333: Nitpick but the indentation here is a little bit off

huwd (approved):

src/components/delete-account/index.njk
```diff
@@ -12,13 +12,26 @@
...
+  <ul class="govuk-list govuk-list--bullet">
+    {% for service in services %}
+
+      {% set locale = ['pages.yourServices.clientRegistry.', env, '.', service.client_id, '.'] | join %}
+        <li id="service-list-item">{{ [locale, 'header'] | join | translate }}</li>
+
+    {% endfor %}
+    </ul>
+    {% if hasGovUkEmailSubscription %}
+    <p class="govuk-body" id="govuk-email-subscription-info">{{ 'pages.deleteAccount.paragraph3' | translate }}</p>
```
danacotoran, line 27: Nitpick: we can tidy up the indentation in this template to make it a little more readable

src/locales/cy/translation.json
divyaparameswaran: The header is used to display content on the delete page.

divyaparameswaran: Yes I think I have added it , but not on ofqual .

PR #860, opened 2023-03-07
GUA-696: tweak header hierarchy and add tests

danacotoran:
A google lighthouse report I ran on the service dashboard has highlighted that on the services dashboard heading levels are not in a sequential order when the user only has accounts/services due to the fact that heading hierarchy skips from `h1` to `h3`  ( the `h2`s “Your accounts“/“Your services“ are not displayed unless the user has both accounts and services on their dashboard).

Screenshot 2023-02-27 at 17 28 44

This introduces some simple logic which changes the heading level depending on context.

It also adds an integration test for **Your services** as the test was missing before.

### DIFF
--- a/src/components/your-services/index.njk
+++ b/src/components/your-services/index.njk
@@ -2,6 +2,8 @@
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
 {% set pageTitleName = 'pages.yourServices.title' | translate %}
 {% set activeNav = 'your-services' %}
+{% set showSecondaryHeadings = servicesList.length and accountsList.length %}
+{% set serviceHeadingLevel =  "h3" if showSecondaryHeadings else "h2" %}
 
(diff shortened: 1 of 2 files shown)


huwd (approved):
Cool! good spot

Was there anything else on the lighthouse report?

danacotoran:
Sorry @huwd saw this comment yesterday and forgot to respond for some reason! 
The main other action on the report was to enable [compression of text-based resources](https://developer.chrome.com/docs/lighthouse/performance/uses-text-compression/?utm_source=lighthouse&utm_medium=devtools) which I'd be keen to implement (even though it's not a particularly frontendy task 😄 ) as there is the potential to reduce payload drastically (but it might be worth discussing beforehand whether that's something we want to do? Not sure whether there could be any potential drawback for instance).

TaherKapasi (approved):

PR #888, opened 2023-03-29
Bump P-memoize from v4.0.1 to v7.1.1 and deprecate commonjs for ESM

huwd:
See commit messages for details.

- Upgrade p-memoise from v4..v7, and fix oidc utils code as a result of changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "libphonenumber-js": "^1.9.51",
         "nunjucks": "^3.2.3",
         "openid-client": "^5.3.2",
-        "p-memoize": "^4.0.1",
+        "p-memoize": "^7.1.1",
         "pino": "^7.0.5",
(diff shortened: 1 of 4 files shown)


huwd:
Re-created from https://github.com/alphagov/di-account-management-frontend/pull/828

huwd:
Rightio, this is a little bigger than I expected 🤔 so in v5 p-memoize makes the jump to requring ESM.

As far as I can tell that would mean removing commonJS throughout our application.

Seems that's a direction of travel for node, so worth looking into, but likely a large piece of work:
https://nodejs.org/api/esm.html

alex9smith:
I'm going to close this PR for now. Like Huw said, this is a largish piece of work and `main` has changed a fair amount since the PR was opened. It'll be easier to start from a clean slate.

PR #890, opened 2023-03-29
Remove yarn

huwd:
We're currently running both npm and yarn, that's silly and I think is confusing our security updates at times, or worse leading to them being resolved in one lockifle but not the one we actually use.

Let's standardise, and I say NPM.

Many of the reasons we chose yarn have now been resolved in NPM (Frozen lockfiles etc).

Open to dissent if folks disagree


huwd:
@mrwilson has rightly pointed out that there are ADR references to why we use yarn.
We should do a quick agreement and get those updated, but I think i agree it doesn't make sense to barge ahead until we've done the policy bit.

huwd:
I'm running a bit ahead of the ADR process here:
https://github.com/alphagov/digital-identity-architecture/pull/389

But after raising at tech leads we had broad agreement, so merging this ahead of the docs catching up:
https://github.com/alphagov/digital-identity-architecture/pull/389

PR #891, opened 2023-03-29
BAU: Update i18next to v22.4.13

huwd:
Seems the breaking change here was about typescript definitions.
But seems internal to them 🤔 

So if this compiles then OK:

See here: https://github.com/i18next/i18next/releases/tag/v22.0.0


huwd:
Closes https://github.com/alphagov/di-account-management-frontend/pull/876

divyaparameswaran (approved):

PR #892, opened 2023-03-29
BAU: Update uuid to v9

huwd:
Breaking changes are listed here:
https://github.com/uuidjs/uuid/blob/main/CHANGELOG.md#900-2022-09-05

Drops a version of node we don't care about (10)

Drops browserside support for browsers some older browser versions. Not a concern here I think


huwd:
Closes: https://github.com/alphagov/di-account-management-frontend/pull/837

divyaparameswaran (approved):

PR #899, opened 2023-04-04
GUA-760: Refactor integration tests

danacotoran:
Update integration tests to use data attribute selectors instead of IDs. 
Remove string checking unless it's relevant to the test (for instance I've left them on the tests dedicated to checking that the specific error messages displayed are correct in different circumstances)

Nothing should change and all the tests should still pass.

https://govukverify.atlassian.net/browse/GUA-760

### DIFF
--- a/test/utils/helpers.ts
+++ b/test/utils/helpers.ts
@@ -1,0 +1,3 @@
+export function testComponent(componentId: string): string {
+  return `[data-test-id='${componentId}']`;
+}
--- a/src/components/change-email/index.njk
+++ b/src/components/change-email/index.njk
@@ -23,6 +23,7 @@
     name: "email",
     value: email,
     errorMessage: {
+        attributes: { "data-test-id": "email-error"},
         text: errors['email'].text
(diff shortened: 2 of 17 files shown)


test/utils/helpers.ts
```diff
@@ -0,0 +1,3 @@
+export function testComponent(componentId: string) {
```
huwd, line 1:
```suggestion
export function testComponent(componentId: string): string {
```

ooh we're missing the return type on this, I'd need to fire this up in vs-code with my typescript tooling to check, but I think it just needs this. It's always gonna return a string, right?

huwd (commented):
Looks great, just need to put a return type definition on the test helper.
I think I've suggested the right thing but would need to double check in my IDE

But really like the approach, nice one :)

test/utils/helpers.ts
danacotoran: Thanks for this @huwd and yes it should always return a string indeed

huwd (approved):

PR #906, opened 2023-05-02
GUA-670 send publicSubjectID to GOV.UK publishing not subjectID

DougsCER:
## Proposed changes
[GUA-670] send publicSubjectID to GOV.UK publishing not subjectID

[GUA-670]: https://govukverify.atlassian.net/browse/GUA-670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/delete-account/delete-account-controller.ts
+++ b/src/components/delete-account/delete-account-controller.ts
@@ -77,7 +77,7 @@ export function deleteAccountPost(
       })
       .catch((err) => {
         req.log.error(
-          `Unable to send delete account notification for:${subjectId}. Error:${err}`
+          `Unable to send delete account notification for:${publicSubjectId}. Error:${err}`
         );
       });
 
--- a/src/components/check-your-email/check-your-email-controller.ts
+++ b/src/components/check-your-email/check-your-email-controller.ts
@@ -29,7 +29,6 @@ export function checkYourEmailPost(
     const {
       email,
       newEmailAddress,
-      subjectId,
       publicSubjectId,
       legacySubjectId,
     } = req.session.user;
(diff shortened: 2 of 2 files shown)


src/components/delete-account/delete-account-controller.ts
```diff
@@ -56,7 +56,7 @@ export function deleteAccountPost(
 
     const DeleteTopicARN = getSNSDeleteTopic();
     try {
-      await service.deleteServiceData(subjectId, DeleteTopicARN);
+      await service.deleteServiceData(publicSubjectId, DeleteTopicARN);
```
divyaparameswaran, line 59: this is for deleting service Data which uses the subject_id . I think the issue was just the logging 
https://github.com/alphagov/di-account-management-frontend/blob/main/src/components/delete-account/delete-account-controller.ts#L80 (which refers to the subject_id instead of public subject_id)

DougsCER: Ah. So my testing did work but i wasn't testing the right thing. 
Glad you understand it.  thank you.   I've updated the PR now

divyaparameswaran: Thank you :) Should we also do it here 
https://github.com/alphagov/di-account-management-frontend/blob/f8638914456e6252ff50da3dc07c6ecaf9eb6ba8/src/components/check-your-email/check-your-email-controller.ts#L64

divyaparameswaran (approved):

PR #908, opened 2023-05-05
GUA-777: Add Security page

danacotoran:
## Proposed changes

We’ve redesigned the ‘settings’ page in preparation for the activity log going live.

We can deploy the new designs now, though. They include:

- changing the title and URL from ‘settings’ to ‘security’
- removing ‘you’re signed in as…'
- moving to a design where the different sections are in boxes
- updating how we play back the user’s phone number

The new designs are in this figma: https://www.figma.com/file/0JbNRTnq5ibRtfg8BMxXDL/Accounts-design-thinking?node-id=1859-41234&t=skBWRYie1VICBikR-4

### What changed
- changing the title and URL from ‘settings’ to ‘security’
- removing ‘you’re signed in as…'
- moving to a design where the different sections are in boxes
- updating how we play back the user’s phone number

### Why did it change
UR suggested representing the page as "Security" makes more sense to users.

### Issue tracking
https://govukverify.atlassian.net/browse/GUA-777


src/components/redirects/redirects-routes.ts
```diff
@@ -21,11 +21,11 @@ router.get(PATH_DATA.THANKS_TXT.url, (_, res) => (
   )
 );
 
-async function redirectManageYourAccount(
+async function redirectToSecurity(
```
huwd, line 24: I think this is triggering some sonarcloud warnings.
Does this need to be an async function?

If so, there may be a conflict, sonarcloud thinks the datatype should be `void` and not `Promise<void>`

src/components/security/index.njk
```diff
@@ -0,0 +1,103 @@
...
+        {% endif %}
+      </div>
+      <div class="summary-list-container">
+        <h2 class="govuk-heading-m">{{ 'pages.security.deleteAccount.heading' | translate }}</h2>
+        <p class="govuk-body">{{ 'pages.security.deleteAccount.info' | translate }}</p>
+        <p class="govuk-body"><a href="/enter-password?type=deleteAccount" class="govuk-link">{{ 'pages.security.deleteAccount.link' | translate }}</a></p>
+      </div>
+    </div>
+  </div>
+{% endblock %}
```
huwd, line 103: If I was being a real pedant... could I ask for a newline at the end of this...

Sorry... only so next time I open and save this my editor doesn't end up adding random new lines to the end 🙈

src/components/redirects/redirects-routes.ts
```diff
@@ -6,9 +6,9 @@ import { requiresAuthMiddleware } from "../../middleware/requires-auth-middlewar
 const router = express.Router();
 
 router.get(
-  PATH_DATA.MANAGE_YOUR_ACCOUNT.url,
+  [PATH_DATA.MANAGE_YOUR_ACCOUNT.url, PATH_DATA.SETTINGS.url],
```
huwd, line 9: Huh, I didn't know this could take an array.

So what are we saying here that if you try to go to either the URLs for Manage your account or settings...
then redirect to settings?

Are we deprecating both? I need to double check the design again...

Cool to know .get() [takes an array](https://expressjs.com/en/api.html#app.get) though, TIL.

src/components/security/security-controller.ts
```diff
@@ -1,14 +1,13 @@
 import { Request, Response } from "express";
-import { redactPhoneNumber } from "../../utils/strings";
 import { getManageGovukEmailsUrl } from "../../config";
 
-export function manageYourAccountGet(req: Request, res: Response): void {
+export function securityGet(req: Request, res: Response): void {
   const data = {
     email: req.session.user.email,
-    phoneNumber: redactPhoneNumber(req.session.user.phoneNumber),
+    phoneNumber: req.session.user.phoneNumber.slice(-4),
```
huwd, line 7: Oh cool, nice one.
So does anything still use that `redactPhoneNumber` function in utils, or can we remove it?

src/components/redirects/redirects-routes.ts
huwd: ah wait, I get this now, because we already had the old Manage your account URL and now we're also deprecating settings, right?

Sorry up to speed now.

huwd (approved):

src/components/redirects/redirects-routes.ts
danacotoran: Yeah that's right, "Settings" used to be "Manage your account", so there was already a redirect in place for that. But now "Settings" is becoming "Security", so we need to ensure both `/manage-your-account` and `/settings` redirect the user to `/security` now.

src/components/security/security-controller.ts
danacotoran: It's still referenced in a few controllers unfortunately! As we are still displaying the phone number in that format in a few places

src/components/redirects/redirects-routes.ts
danacotoran: I don't _think_ it needs to be async so I changed to a regular one. The weird thing is I didn't do anything about the other Sonarcloud warning, yet it disappeared.

huwd (approved):
Still a bit concerned about the sonarcloud issue, but if the linting passes 🤔 

really makes me think sonarcloud's value is questionable here - I don't like flaky CI

PR #912, opened 2023-05-22
FIX: Phone number is not guarenteed to exist

huwd:
Phone number might not exist, we can't perform slice on undefined.
Which is leading to errors.


huwd:
We should follow this up with another test case

danacotoran (approved):
thanks for the fix, sorry we missed it

PR #913, opened 2023-05-22
FIX: Harmonise treatment of 00XX international phone number format with Auth

DTaylorGDS:
## Proposed changes
- Accept international phone numbers in the format 00XX
- Previously only E164 format numbers (e.g. +33) were accepted
- User research showed that this was unhelpful to end users
- We therefore already loosened this on the sign-in side (auth)
- Now we should harmonise the approach to cover account management too
- Whilst making this change, I noticed that we had different length tolerance in validation. I have harmonised to the longer value (auth's): max length of international numbers 26 chars
- Corresponding change on auth side can be found in commit 044ba5e906afa568bb073d05bd9a8e902d220928 in di-authentication-frontend

### What changed
- The 'sort' of international number format that should be accepted in front end validation. Before this would be rejected but after this PR it will be accepted:
image

### Why did it change
- User research suggested that we were being too restrictive: https://gds.slack.com/archives/C01TN7NC951/p1683644537435549
- We therefore made a change for the sign-in journey: https://github.com/alphagov/di-authentication-frontend/pull/1019
- This would now keep account management behaviour consistent with that change

### Issue tracking
- Mirrors change on auth team side: https://github.com/alphagov/di-authentication-frontend/pull/1019

### Other considerations
- [x] Demo to a BA, TA, and the team - no direct UI change, but new behaviour has been requested by Content Designer
- [x] Update [README](./blob/main/README.md) with any new instructions or tasks - Not needed


huwd (commented):
Looks good to me but might ask @danacotoran to take a look as our frontender and @DougsCER to take a look as someone who worked in this patch most recently.

Do you have an ideal window to get this merged @DTaylorGDS?

huwd (approved):

PR #915, opened 2023-06-09
BAU : Add th resend link and phone number link to the view as hidden …

divyaparameswaran:


DougsCER (approved):

src/components/check-your-phone/index.njk
```diff
@@ -21,6 +21,8 @@
     <form action="/check-your-phone" method="post" novalidate="novalidate">
         <input type="hidden" name="_csrf" value="{{csrfToken}}"/>
         <input type="hidden" name="phoneNumber" value="{{phoneNumber}}"/>
+        <input type="hidden" name="resendCodeLink" value="{{resendCodeLink}}"/>
+        <input type="hidden" name="changePhoneNumberLink" value="{{changePhoneNumberLink}}"/>
```
DougsCER, line 25: should the code be commented to explain WHY those fields are present and hidden?
It's obvious what the code is doing. but not why. 
code should be self explanatory what it's doing, comments are helpful to explain why it needs to do that thing

src/components/check-your-phone/index.njk
```diff
@@ -21,6 +21,9 @@
     <form action="/check-your-phone" method="post" novalidate="novalidate">
         <input type="hidden" name="_csrf" value="{{csrfToken}}"/>
         <input type="hidden" name="phoneNumber" value="{{phoneNumber}}"/>
+         {# These feilds are needed when the page is re-rendered once an error has occured #}
```
huwd, line 24:
```suggestion
         {# These fields are needed when the page is re-rendered once an error has occurred #}
```

Could correct this here, I'd also be up for moving the explanation out of a comment and into a commit message.
But feels like this is one where we want to move quickly and tidy up later.

So go with whatever makes sense to you right now!

huwd (approved):

PR #927, opened 2023-07-04
GUA-874: Use stub account management API in build

alex9smith:
Picks up the SSM parameter exported from the backend stack for the new stub API endpoint added in https://github.com/alphagov/di-account-management-backend/pull/123.

This has to be a bit clunky with the condition and if clause in the task definition because Cloudformation doesn't let you resolve SSM parameters in the mapping section.


alex9smith:
This is safe to go now. If there are problems with the stub in build that block something else we can always open another PR to use the real API again.

DougsCER (approved):

PR #933, opened 2023-07-11
OLH-932: Add apprenticeship service

danacotoran:
Apprenticeship service is onboarding on July 13th. 

This adds it to the locale files, services allow list and the list of services we display in the information box on the account.

https://govukverify.atlassian.net/browse/DOT-51
https://govukverify.atlassian.net/browse/OLH-932

## Testing

Following the steps in [Adding a New client](https://docs.google.com/document/d/1qMUACpf4IuaMGEQzPFSA6ZoIkgbwMMuVk2E9qKHj1PQ/edit).
Recorded demo posted [here](https://gds.slack.com/archives/C01H13E9YP8/p1689094331301769).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -122,10 +122,12 @@ export const getAllowedAccountListClientIDs: string[] = [
   "CEr97IZfEPQFgBxq8QNcM8LFxw4",
   "TGygWFxGDNn8ItyaecWCopqIX3s",
   "pDqO7_Hu-pq5wam5I4MlURXrv5k",
+  "x3F_Iu0LgqJpegY5ni0QSB0uezw",
   "gov-uk",
   "lite",
   "ofqual",
   "modernSlavery",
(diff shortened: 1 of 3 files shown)


alex9smith (approved):
I see that some services (including this one) have the same client ID in integration and production, while others have different ones. Let's not block this PR from going live because I see we don't have the integration one on the ticket, but is this something we need to sort out?

danacotoran:
Ohh that's a good point @alex9smith. To be honest I kind of assumed that it should be the same. I wonder if there is some reason some of them have different IDs in different environments

alex9smith:
As far as I know they're an auto-generated random string, so I'd expect them to be different in each environment unless we've gone out of our way to keep them the same. 

I'll ask if the Auth team can find us the rest of the integration ones and see about getting the ticket template updated to include that too.

PR #977, opened 2023-08-08
OLH-950: Update link to GOV.UK email subscriptions

alex9smith:
### What changed

Add a URL parameter to all links to GOV.UK email subscriptions.

Also correct the environments - GOV.UK's integration environment uses our staging one, and their staging uses our integration. 

### Why did it change

The GOV.UK email subscriptions service currently has two ways to authenticate a user. Adding this URL parameter to links from the GOV.UK One Login home allows the email service to know that that user should be authenticated through GOV.UK One Login.

This means the user can skip a step when signing into the email service as it no longer needs to work out which authentication method to use.

## Testing

I've run the app locally and checked the URL parameter was present.

### DIFF
--- a/src/locales/cy/translation.json
+++ b/src/locales/cy/translation.json
@@ -207,7 +207,7 @@
             },
             {
               "text": "Tanysgrifiadau e-byst GOV.UK",
-              "href": "https://www.gov.uk/email/manage"
+              "href": "https://www.gov.uk/email/manage?from=your-services"
             },
             {
               "text": "LITE (Trwyddedu ar gyfer Masnach a Menter Ryngwladol)",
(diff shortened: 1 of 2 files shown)


danacotoran:
The list of services under `clientRegistry` in the locale files is used not only on the Your services page but also the Sign in history page and the Delete your account page. 
Just to confirm @alex9smith, is it our intention that the link has the `?page=your-services` parameter in all these places, and is there no need currently to differentiate which page on the One Login account the user has arrived from when they land on the GOVUK email subscription service?

DougsCER (approved):

alex9smith:
@danacotoran yes, all users clicking a link to GOV.UK email subscriptions from our app should be using GOV.UK One Login to authenticate. So we want to add the parameters to links in all the possible places and there's only one entry point to the service at the moment.

danacotoran:
Cool, thanks for confirming @alex9smith. Wonder if the parameter should have been named something different then, like `?from=one-login` but... oh well 😄 
Sounds like it's only used programmatically so, as long as it's documented and doesn't confuse the developers, I suppose!

PR #1038, opened 2023-09-13
Olh 1089 handle auth for deeplinks

divyaparameswaran:
## Proposed changes
When we link to the /security page from an external services it redirects to you session time put as the Session does not exist. To avoid this , a check is included to see if session exists , if not then it allows users to Sign in and then redirects to the page . 
### What changed

Included a check to see if session is undefined and add logic to create the session and redirect to the Auth Sign in screen .
### Why did it change
The change was in response to an incident : https://govukverify.atlassian.net/browse/INCIDEN-158

Steps for testing are on the ticket : https://govukverify.atlassian.net/browse/OLH-1089


src/middleware/requires-auth-middleware.ts
```diff
@@ -18,3 +28,22 @@ export function requiresAuthMiddleware(
     next();
   }
 }
+
+async function startGet(req: Request, res: Response): Promise<void> {
```
alex9smith, line 32: I think `startGet` is a confusing name for this function here, could we call it something like `redirectToLogIn` instead?

alex9smith (commented):
This is a good start, but I don't think it solves the whole problem. We also need to make sure the user comes back to the same page they were trying to get to after they sign in. 

I think we need to add the current URL to the user's session here, then in the callback controller we can modify [this line](https://github.com/alphagov/di-account-management-frontend/blob/main/src/components/oidc-callback/call-back-controller.ts#L55) to check for that URL in their session. Let me know if you want me to explain more / push up some code to demonstrate. 

Could you also add a test for this behaviour? The middleware tests are in `di-account-management-frontend/test/unit/middleware`.

src/middleware/requires-auth-middleware.ts
divyaparameswaran: Ah yes sure

alex9smith (approved):
Looks good to me! Thanks for the updates

PR #1050, opened 2023-09-26
GA4 page view

danacotoran:
## Proposed changes
This ticket covers the MVP requirement for GA4 On Page Load data layer push.
https://govukverify.atlassian.net/browse/OLH-974

### What changed
Refactor existing cookie and analytics code to separate them. 
Remove some unused cookie related functions. 
Separate shared analytics related code from UA/GA4 code.
Introduce more consistent namespacing.
Add GA4 page view event.

This does not include level 2 taxonomy which is covered by another ticket.

### Why did it change
UA is going away.

## Testing
This has already been deployed to dev where it's been tested by a PA who confirmed that the events were showing up in the GA interface as expected.
I have already tested it manually with the feature flag switched off and on. 

## How to review
Besides checking that the requirements outlined in the JIRA ticket have been met, it would be good if we could check the following:
- cookie banner still works the same way as it worked before
- UA still works the same way it worked before
- the `supportGa4` feature flag works as expected
- there is no potential for Javascript errors that would break the rest of the code, if they occurred


danacotoran:
@gtvj would you be able to have a look at this when you get a minute? Thanks! 🙏

PR #1057, opened 2023-10-04
OLH-1114 Extract fromUrl from request param or session for Contact Centre Page

DougsCER:
## Proposed changes
`[OLH-1114] Extract fromUrl from request param or session for Contact Centre Page`

### What changed
Check query param and session for a valil fromUrl. If not present then return an error page and log the error. 

- [/] No environment variables or secrets were added or changed

## Testing
Unit tests. 
Deployed to dev and saw page loading in all scenarios. Can't yet test the link to contact form because the link will be created on a different ticket.

[OLH-1114]: https://govukverify.atlassian.net/browse/OLH-1114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
+++ b/src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
@@ -1,24 +1,48 @@
 import { Request, Response } from "express";
+import pino from "pino";
+import { isValidUrl } from "./../../utils/strings";
+<<<<<<< HEAD
 import {
   supportWebchatContact,
   supportPhoneContact,
   showContactGuidance,
 } from "../../config";
 
+const CONTACT_ONE_LOGIN_TEMPLATE = "contact-govuk-one-login/index.njk";
+const logger = pino();
+
 export function contactGet(req: Request, res: Response): void {
   const isAuthenticated = req.session?.user?.isAuthenticated;
   let isLoggedOut = req.cookies?.lo;
+  const fromURL = getFromUrlAndSaveIt(req);
 
   if (typeof(isLoggedOut) === 'string') {
     isLoggedOut = JSON.parse(isLoggedOut)
+  }
+  
+  if (!fromURL) {
+    logger.info(
+      "Request to contact-govuk-one-login page did not contain a valid fromURL in the request or session"
+    );
   }
 
   const data = {
     contactWebchatEnabled: supportWebchatContact(),
     contactPhoneEnabled: supportPhoneContact(),
     showContactGuidance: showContactGuidance(),
-    showSignOut: isAuthenticated && !isLoggedOut
+    showSignOut: isAuthenticated && !isLoggedOut,
+    fromURL
   };
   
   res.render("contact-govuk-one-login/index.njk", data);
 }
+
+const getFromUrlAndSaveIt = (request: Request): string => {
+  const fromURLFromRequest = request.query.fromURL as string;
+  if (isValidUrl(fromURLFromRequest)) {
+    request.session.fromURL = fromURLFromRequest;
+    return fromURLFromRequest;
+  }
+  const fromURLFromSession = request.session.fromURL;
+  return fromURLFromSession;
+};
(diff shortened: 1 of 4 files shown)


src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -1,7 +1,30 @@
...
-
-  res.render("contact-govuk-one-login/index.njk", data);
+  const fromUrl = getFromUrlAndSaveIt(req);
+  if (fromUrl) {
+    res.render(CONTACT_ONE_LOGIN_TEMPLATE, data);
+  } else {
+    req.log.error(
+      `request and session contain no valid fromUrl`
+    );
+    res.render(NOT_FOUND_TEMPLATE);
```
DougsCER, line 16: I'd appreciate your thoughts on the best approach here.

alex9smith: Hmm good point. 

This is just a web page on the internet, so it's entirely possible people end up on it without a URL param (eg. via a Google search). I think we still need to let those people contact us, so if there's no `fromUrl` in the params or in the session then we should not include that param in the link off to the contact form. I'll check with Auth that they can still handle that case.

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -1,7 +1,29 @@
...
-export function contactGet(req: Request, res: Response): void {
-  const data = {};
+const CONTACT_ONE_LOGIN_TEMPLATE = "contact-govuk-one-login/index.njk";
+const logger = pino();
 
-  res.render("contact-govuk-one-login/index.njk", data);
+export function contactGet(req: Request, res: Response): void {
+  const fromUrl = getFromUrlAndSaveIt(req);
+  if (!fromUrl) {
+    logger.error(
```
alex9smith, line 11:
```suggestion
    logger.info(
```
Now we're not showing users the 404 page when the param isn't present, can you make this INFO level? I don't want to be tripping any alarms when things are working as we expect.

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -1,7 +1,29 @@
...
+    );
+  }
+  const data = {
+    fromUrl: fromUrl,
+  };
+  res.render(CONTACT_ONE_LOGIN_TEMPLATE, data);
 }
+
+const getFromUrlAndSaveIt = (request: Request): string => {
+  const fromUrlFromRequest = request.query.fromUrl as string;
```
alex9smith, line 22:
```suggestion
  const fromUrlFromRequest = request.query.fromURL as string;
```
We've asked teams to call the parameter `fromURL` (note the capitals). I don't think we need to bother changing the name elsewhere in our code

alex9smith (changes requested):
Just a couple of small changes 👍🏻

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
DougsCER: I would argue it's still an error that we are gracefully handling. But unexpected and shouldn't happen. But i can see your point so will update it

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -1,7 +1,29 @@
...
-  res.render("contact-govuk-one-login/index.njk", data);
+export function contactGet(req: Request, res: Response): void {
+  const fromURL = getFromUrlAndSaveIt(req);
+  if (!fromURL) {
+    logger.info(
+      "Request to contact-govuk-one-login page did not contain a valid fromURL in the request or session"
+    );
+  }
+  const data = {
+    fromURL: fromURL,
```
aenciso, line 16: I think this can be simplified to 
`{ fromURL }`

DougsCER: Resolved this in the next PR 
https://github.com/alphagov/di-account-management-frontend/pull/1058

alex9smith (approved):

PR #1059, opened 2023-10-06
OLH-1113: Contact centre triage page

danacotoran:
## Proposed changes
Build a static version of the triage page using the content and designs from [the prototype](https://di-contact-centre-2dd92a1953cd.herokuapp.com/exploratory-concepts/v1/contact-triage#). 

We don’t yet have final content for the page or Welsh translations so the page will be all in English for now. 
The links on the page aren't working yet either as it's not fully clear what they should be.

The value of the reference code is also hard coded for now. 

For now the webchat section, phone contact section, and guidance section are all behind flags.
The page without guidance links, webchat or phone contact:
Screenshot 2023-10-06 at 12 52 23

The page with everything switched on:
localhost_6001_contact-govuk-one-login

Another detail included in the PR is that a **Sign out** link displays in the header when the user is authenticated with OL.

Screenshot 2023-10-06 at 13 00 44

## How to review

- Check the content is correct (already demo'ed to UCD and they approved
- Check the flags all work
- Check that the changes to how the account navigation is displayed, and the new Sign out link added to the header for this page, aren't breaking anything


src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -1,7 +1,24 @@
 import { Request, Response } from "express";
+import {
+  supportWebchatContact,
+  supportPhoneContact,
+  showContactGuidance,
+} from "../../config";
 
 export function contactGet(req: Request, res: Response): void {
-  const data = {};
+  const isAuthenticated = req.session?.user?.isAuthenticated;
```
DougsCER, line 9: @danacotoran  have you tried this out?
As the RequiresAuthMiddleware is not called (in the root file) it's possible some people may be sent to this page from another system (where they are logged in) but we don't have the isAuthenticated field set in the session

danacotoran: @DougsCER I tested it locally and it worked as expected (as in, I could see the "Sign out" link only when I was authenticated with my OL) but you're probably better positioned to identify the risks/issues to do with this approach. So if you would advise that this isn't a good idea, the UCD folks said this was a "good to have" so we could probably go without it.

DougsCER: Berne thinking about this over the weekend. 
This approach will correctly show the sign-out link for those who have visited a page on AMF that requires authentication. 
Users that have signed in on another service then follow the link here won't be shown the sign-out link. This would be nice to do but not simple.  So I think we should progress as-is and consider if there is an improvement we can do in future to show the link for people who have signed in on another service.

DougsCER (approved):

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
alex9smith: I agree - let's carry on as is and I'll make a tech debt story to remember this

PR #1066, opened 2023-10-12
Fix link to sign your mortgate deed.

andrew-moores:
## Proposed changes

OLH-1145 - Fix link to sign your mortgage deed service
### What changed
The url has been corrected.

### Why did it change
The old url was incorrect and led to a page stating the service was unavailable.

## Testing

Built and tested locally.

### DIFF
--- a/src/locales/cy/translation.json
+++ b/src/locales/cy/translation.json
@@ -231,7 +231,7 @@
             },
             {
               "text": "Llofnodwch eich gweithred morgais",
-              "href": "https://sign-your-mortgage-deed.landregistry.gov.uk/interceptor"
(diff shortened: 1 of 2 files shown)


DougsCER (approved):

src/locales/cy/translation.json
```diff
@@ -806,7 +806,7 @@
       "mortgageDeed": {
```
DougsCER, line 806: Am i right this is appearing twice in the dev list?

andrew-moores: I can only see one entry in the dev list, the entry at line 806 is in the "local" list.

DougsCER: nope that's in 'local'
my bad. 
this is all good

PR #1067, opened 2023-10-12
OLH-1117 Build welsh language toggle

aenciso:
## Proposed changes

[OLH-1117] Add a Welsh language toggle to the triage page to allow users to easily switch languages.

 - [ ] Added env variables to docker file

## Testing

Tested locally switching between languages by clicking on the links
Tested by landing on the page without any links
Tested by having parameters already in the URL

[OLH-1117]: https://govukverify.atlassian.net/browse/OLH-1117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -300,37 +300,36 @@ $nav-active-border-thickness: 5px;
     display: block;
     clear: both;
   }
+}
 
-  .language-select__list {
-    border-bottom: 1px solid $govuk-border-colour;
-    margin-bottom: 0;
-    float: right;
-    text-align: right;
-    @include govuk-media-query($until: tablet) {
-      float: none;
-      text-align: left;
-    }
+.language-select__list {
+  border-bottom: 1px solid $govuk-border-colour;
+  margin-bottom: 0;
+  float: right;
+  text-align: right;
+  @include govuk-media-query($until: tablet) {
+    float: none;
+    text-align: left;
   }
+}
 
-  .language-select__link {
-    padding: govuk-spacing(1);
-  }
+.language-select__link {
+  padding: govuk-spacing(1);
+}
 
-  .language-select__list-item {
-    @include govuk-font($size: 16);
-    display: inline-block;
-    margin-right: 0;
-    border-bottom: 0;
-  }
+.language-select__list-item {
+  @include govuk-font($size: 16);
+  display: inline-block;
+  margin-right: 0;
+  border-bottom: 0;
+}
 
-  .language-select__list-item [aria-current] {
-    padding: govuk-spacing(1);
-  }
+.language-select__list-item [aria-current] {
+  padding: govuk-spacing(1);
+}
 
-  .language-select__list-item + .language-select__list-item::before {
-    content: "|";
-    color: $govuk-border-colour;
-  }
-
+.language-select__list-item + .language-select__list-item::before {
+  content: "|";
+  color: $govuk-border-colour;
 }
 
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -314,7 +314,7 @@ $nav-active-border-thickness: 5px;
 
   .language-select__link {
     @include govuk-font($size: 16);
-    padding: 0.3125em;
+    padding: govuk-spacing(1);
 
     &:visited {
       color: $govuk-link-colour;
@@ -336,7 +336,7 @@ $nav-active-border-thickness: 5px;
   }
 
   .language-select__list-item [aria-current] {
-    padding: 0.3125em;
+    padding: govuk-spacing(1);
   }
 
   .language-select__list-item + .language-select__list-item::before {
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -47,8 +47,8 @@ $govuk-new-link-styles: true;
   }
 
   @include govuk-media-query($from: "tablet") {
-    .govuk-summary-list__key, 
-    .govuk-summary-list__value, 
+    .govuk-summary-list__key,
+    .govuk-summary-list__value,
     .govuk-summary-list__actions {
       padding-top: govuk-spacing(4);
       padding-bottom: govuk-spacing(4);
@@ -70,8 +70,8 @@ $govuk-new-link-styles: true;
   }
 
   @include govuk-media-query($from: "tablet") {
-    .govuk-summary-list__key, 
-    .govuk-summary-list__value, 
+    .govuk-summary-list__key,
+    .govuk-summary-list__value,
     .govuk-summary-list__actions {
       padding-top: govuk-spacing(4);
       padding-bottom: govuk-spacing(4);
@@ -207,7 +207,7 @@ $nav-active-border-thickness: 5px;
   font-weight: bold;
   padding: $nav-link-padding - $nav-link-outline-thickness 0 $nav-link-padding;
   margin-top: $nav-link-outline-thickness;
-  
+
   &:not(:hover) {
     text-decoration: none;
   }
@@ -291,3 +291,58 @@ $nav-active-border-thickness: 5px;
   margin: govuk-spacing(2) 0;
   @include govuk-font($size: 48, $weight: bold);
 }
+
+.language-select {
+  margin: govuk-spacing(2) 0;
+
+  &::after {
+    content: "";
+    display: block;
+    clear: both;
+  }
+
+  .language-select__list {
+    border-bottom: 1px solid $govuk-border-colour;
+    margin-bottom: 0;
+    float: right;
+    text-align: right;
+    @include govuk-media-query($until: tablet) {
+      float: none;
+      text-align: left;
+    }
+  }
+
+  .language-select__link {
+    @include govuk-font($size: 16);
+    padding: 0.3125em;
+
+    &:visited {
+      color: $govuk-link-colour;
+
+      &:hover {
+        color: $govuk-link-hover-colour;
+      }
+
+      &:focus {
+        color: $govuk-focus-text-colour;
+      }
+    }
+  }
+
+  .language-select__list-item {
+    display: inline-block;
+    margin-right: 0;
+    border-bottom: 0;
+  }
+
+  .language-select__list-item [aria-current] {
+    padding: 0.3125em;
+  }
+
+  .language-select__list-item + .language-select__list-item::before {
+    content: "|";
+    color: $govuk-border-colour;
+  }
+
+}
+
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -313,11 +313,11 @@ $nav-active-border-thickness: 5px;
   }
 
   .language-select__link {
-    @include govuk-font($size: 16);
     padding: govuk-spacing(1);
   }
 
   .language-select__list-item {
+    @include govuk-font($size: 16);
     display: inline-block;
     margin-right: 0;
     border-bottom: 0;
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -315,18 +315,6 @@ $nav-active-border-thickness: 5px;
   .language-select__link {
     @include govuk-font($size: 16);
     padding: govuk-spacing(1);
-
-    &:visited {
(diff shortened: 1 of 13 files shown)


docker-compose.yml
```diff
@@ -51,6 +51,13 @@ services:
       - SUPPORT_ACTIVITY_LOG=${SUPPORT_ACTIVITY_LOG}
       - DEBUG=express-session
       - LOCALSTACK_HOSTNAME=host.docker.internal
+      - SUPPORT_TRIAGE_PAGE=${SUPPORT_TRIAGE_PAGE}
```
DougsCER, line 54: I don't think these need to be in the docker compose if they are in your local .env file

aenciso: You're right yes. Changes reverted.

src/assets/scss/application.scss
```diff
@@ -290,4 +290,38 @@ $nav-active-border-thickness: 5px;
...
   margin: govuk-spacing(2) 0;
   @include govuk-font($size: 48, $weight: bold);
-}
\ No newline at end of file
+}
+
+.language-select {
+  float: right;
+  border-bottom: 1px solid $govuk-border-colour;
+  .govuk-list {
```
danacotoran, line 298: In our component model, we try to adhere to a BEM way of writing CSS instead of relying on nesting selectors. 

If you need to style the unordered list element, instead of relying on specificity to overwrite existing rules, we would instead assign another class to `ul` in addition to `govuk-list`. So the `ul` becomes `<ul class="govuk-list language-select__list">`. The links become `<a class="{{insert govuk link classes here}} language-select__link">`. And so on.

So then the SCSS can be changed to:

```
.language-select {
 ... styles specific to the NAV container...
}

.language-select__list {
... styles that are specific to the UL element only
}

.language-select__link {
... styles that are specific to the anchor tag element only
}
```

and so on.

Let me know if any of that doesn't make sense.

src/assets/scss/application.scss
```diff
@@ -290,4 +290,38 @@ $nav-active-border-thickness: 5px;
   display: block;
   margin: govuk-spacing(2) 0;
   @include govuk-font($size: 48, $weight: bold);
-}
\ No newline at end of file
+}
+
+.language-select {
+  float: right;
+  border-bottom: 1px solid $govuk-border-colour;
```
danacotoran, line 297: The component in the design which is linked to in the JIRA ticket does not have a border on the bottom. Obviously you have added this deliberately but as there is a discrepancy I just wanted to make sure which design is the correct one: border or no border? 🤔

src/components/common/layout/languageToggle.njk
```diff
@@ -0,0 +1,18 @@
+<nav class="language-select" aria-label="Choose a language">
+    <ul class="govuk-list">
+        {% for lang in ['en', 'cy'] %}
+            <li class="govuk-header__navigation-item">
```
danacotoran, line 4: I'm interested in the reasoning behind the choice to use the `govuk-header__navigation-item` class. 

Seeing as this class is actually specific to the navigational links in the Design System [header component](https://design-system.service.gov.uk/components/header/), which is unrelated to this component, I would suggest that relying on this class could lead to unintended consequences in the future (it has actually already led to unintended consequences – you have had to over-write some of the styles that this class comes with 😄). 
It might be best to remove this class from the `li`, opting to use a BEM style class such as `language-select__list-item` instead.

src/assets/scss/application.scss
```diff
@@ -290,4 +290,38 @@ $nav-active-border-thickness: 5px;
   display: block;
   margin: govuk-spacing(2) 0;
   @include govuk-font($size: 48, $weight: bold);
-}
\ No newline at end of file
+}
+
+.language-select {
+  float: right;
```
danacotoran, line 296: I would probably choose to avoid using float here (alternatively, clear the float) because it is causing some layout issues on smaller screens where it's almost going in line with the heading element instead of staying above it.

src/assets/scss/application.scss
```diff
@@ -290,4 +290,38 @@ $nav-active-border-thickness: 5px;
...
+  }
+
+  .govuk-header__navigation-item {
+    display: inline-block;
+    margin-right: 0;
+    border-bottom: 0;
+  }
+
+  .govuk-header__navigation-item + .govuk-header__navigation-item::before {
+    content: "|";
```
danacotoran, line 324: Nice solution

src/components/common/layout/languageToggle.njk
```diff
@@ -0,0 +1,18 @@
...
+                {% set translationKey = 'general.languageToggle.' + lang + '.linkText' %}
+                {% if language == lang %}
+                    <span class="govuk-!-font-size-16" aria-current="true">{{ translationKey | translate }}</span>
+                {% else %}
+                    <a aria-current="page"
```
danacotoran, line 9: I might be getting myself confused as this is a part of ARIA I'm not super familiar with but why should non-active items have `aria-current="page"`? 🤔 

The guidance on `aria-current` seems to indicate that a value of `page` should represent that we are currently on the current page within a set of pages.

On the other hand a value of `true` seems a bit more general: represents the current item within a set.

Based on these descriptions I would probably lean towards using "true" as opposed to "page" for the active element. But whether it's "true" or it's "page", I think that only the currently selected option should have the `aria-current` attribute, where the non-selected option(s) shouldn't have `aria-current` at all (or have the attribute, but with a value of "false")

danacotoran (commented):
Hi @aenciso, thank you for your work on this. 
I can't comment on the backend parts of the work but I had some comments/questions around some of the frontend. Do let me know if anything doesn't make sense!

src/locales/cy/translation.json
```diff
@@ -44,6 +44,14 @@
       "english": "Saesneg",
       "welsh": "Cymraeg"
     },
+    "languageToggle": {
+      "en": {
+        "linkText": "Saesneg"
```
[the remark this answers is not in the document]
aenciso: It makes sense. Let me update that.

src/components/common/layout/languageToggle.njk
danacotoran: That's not quite what I was questioning with my comment. I could have written it more clearly.

I think using a span as the active item is fine. The other existing examples of this component use span – not to say that we should indiscriminately copy what other teams in government have done, just saying that there is precedent, and if teams like HMRC and GOVUK are using it in their live service, it's _probably_ okay. 

I also think adding `aria-current="true"` to it is a good idea as it represents the currently active item within a set of items.

What I am not so sure about is why/if the un-selected language should have `aria-current` at all given that it's not the current item.

src/assets/scss/application.scss
aenciso: It was found at a later stage (UCD huddle) that the language component should have a bottom border, that's why it was added.

src/assets/scss/application.scss
aenciso: All done and sorted

src/components/common/layout/languageToggle.njk
aenciso: It was a left over from hmrc component I forgot to replace. They're using in it in a different layout so for them makes sense but not for us. I have updated it accordingly.

src/assets/scss/application.scss
aenciso: Thanks

src/assets/scss/application.scss
aenciso: I've added the clear but left the float just because we need the bottom-border to be the same size of the navigation otherwise it will span the whole container width.
I have tested it on different screen resolutions and it looks fine.

andrew-moores (approved):

src/assets/scss/application.scss
```diff
@@ -290,4 +290,59 @@ $nav-active-border-thickness: 5px;
...
+.language-select {
+  margin: govuk-spacing(2) 0;
+
+  &::after {
+    content: "";
+    display: block;
+    clear: both;
+  }
+
+  .language-select__list {
```
danacotoran, line 304: Thanks for re-writing the classes here! Is there a reason why `.language-select__list` should be nested under `.language-select` in the scss? Is there a need for this higher level of specificity?

src/assets/scss/application.scss
```diff
@@ -290,4 +290,59 @@ $nav-active-border-thickness: 5px;
...
+  }
+
+  .language-select__list-item {
+    display: inline-block;
+    margin-right: 0;
+    border-bottom: 0;
+  }
+
+  .language-select__list-item [aria-current] {
+    padding: 0.3125em;
```
danacotoran, line 339: Interested in this `0.3125em` value! "Seems a little awkward. 
Is this something that we cannot achieve with standard `govuk-spacing`?
Is there a reason this component should be special and should not use standard spacing, like the [GOV.UK version](https://components.publishing.service.gov.uk/component-guide/translation_nav) does?

src/assets/scss/application.scss
```diff
@@ -290,4 +290,59 @@ $nav-active-border-thickness: 5px;
...
+  }
+
+  .language-select__link {
+    @include govuk-font($size: 16);
+    padding: 0.3125em;
+
+    &:visited {
+      color: $govuk-link-colour;
+
+      &:hover {
```
danacotoran, line 322: Is this rule (and the `&:focus` rule below) necessary and why?

src/assets/scss/application.scss
```diff
@@ -290,4 +290,59 @@ $nav-active-border-thickness: 5px;
...
+      float: none;
+      text-align: left;
+    }
+  }
+
+  .language-select__link {
+    @include govuk-font($size: 16);
+    padding: 0.3125em;
+
+    &:visited {
```
danacotoran, line 319: Seeing as the `govuk-link--no-visited-state` modifier class has been used on the link element, is this rule necessary and why?

src/assets/scss/application.scss
```diff
@@ -290,4 +290,59 @@ $nav-active-border-thickness: 5px;
...
+    float: right;
+    text-align: right;
+    @include govuk-media-query($until: tablet) {
+      float: none;
+      text-align: left;
+    }
+  }
+
+  .language-select__link {
+    @include govuk-font($size: 16);
```
danacotoran, line 316: The font size is set on this class using the `govuk-font` mixin but for the current language span element, the govuk font override class is used. Is there a reason for the difference?

danacotoran (commented):
Hi @aenciso thanks for making some changes, the component looks good. 
I have a few questions around your code still, hope they make sense to you.

src/assets/scss/application.scss
aenciso: Totally agree, this is a weird one. This is the value from the prototype. Not sure why they have spacing though but I have just changed it with the govuk-spacing equivalent.

src/assets/scss/application.scss
aenciso: removed

src/assets/scss/application.scss
aenciso: removed

src/assets/scss/application.scss
aenciso: Fixed

src/assets/scss/application.scss
danacotoran: We tend to avoid nesting unless necessary. The times we usually rely on nesting are with modifiers and on the rare occasion we might need to override some different rule so we need that level of specificity. 

In the `.account-navigation__link` example for instance, it has some rules that are specific for the "default" state of the link. But account nav links can also have an "active" state with slightly different styling hence the 
```
.account-navigation__link {
  ... default styles for .account-navigation__link...

  .account-navigation__list-item--active & {
    ...styles to apply "active" state to .account-navigation__link ...
  }
}
```
(not 100% sure why the active modifier was set to the link element's parent instead of the link itself in this particular instance, probably because there were tweaks in the styles of both elements for  the "active" state, but I hope the point makes sense anyway)

aenciso: It makes sense, yes. I have removed all nesting.

PR #1071, opened 2023-10-16
OLH-1168: Put the webchat widget on the triage page

alex9smith:
### What changed

Put the webchat widget on the triage page behind the existing feature flag. The supplier's docs for the webchat ask us to put it in the `<head>` section, but this is the same implementation as on the previous webchat demo page - it works fine in the `<body>` and is much easier to implement in our templates.

### Related links

See #1052 and #1056 for the implementation of the webchat demo page.

- [x] Changed an environment variable

## Testing

I've run this locally and verified the webchat opens when I click the widget. ~~Note that the supplier hasn't added any of our domains to their allowlist yet, so this will only work when running the app locally.~~ They've now added our domains to the allowlist so this will work up to staging.

image

### DIFF
--- a/src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
+++ b/src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
@@ -2,6 +2,7 @@ import { Request, Response } from "express";
 import { logger } from "../../utils/logger";
 import { isSafeString, isValidUrl } from "./../../utils/strings";
 import {
+  getWebchatUrl,
   supportWebchatContact,
   supportPhoneContact,
   showContactGuidance,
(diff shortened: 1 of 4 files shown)


src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -35,6 +36,7 @@ export function contactGet(req: Request, res: Response): void {
     showSignOut: isAuthenticated && !isLoggedOut,
     referenceCode,
     contactEmailServiceUrl: contactEmailServiceUrl,
+    webchatSource: getWebchatUrl(),
```
DougsCER, line 39: does contactWebchatEnabled need adding to the data?

alex9smith: It's already there, but it's not shown in this short diff because we have so many variables in the data for this page

andrew-moores (approved):

DougsCER (approved):

PR #1075, opened 2023-10-17
OLH-1165: Add an example client side JS test

alex9smith:
### What changed

Add an example test for client side JS.

There's a new npm script (`npm run test:js`) that runs mocha looking for *.test.js files in the assets directory.

The example tests use the same `chai` and `mocha` runner and framework that the Typescript tests use. We need to use CommonJS syntax rather than the `import {expect} from "chai"` ES6 syntax, but that's a difference we can manage.

I've not yet added this script to the list of ones which are run in `npm run test` - we can do that once we have some more substantial tests.

### Why did it change

We need to have unit tests for our client side JS before we do any significant work on either the webchat or GA4.

## Testing

Run the new tests with `npm run test:js`

## How to review

I've done very little work with client side JS. This looks sensible to me and the [Chai docs](https://www.chaijs.com/guide/plugins/) suggest that it's capable of managing client side JS, but this is a very simple example. We should consider if it'll work with the module system the rest of our client side code uses.

@andrew-moores I added a test for a simple method in the cookies module to illustrate how we can test our client side code.


alex9smith:
Nice! This is even simpler than I was hoping it could be.

Turns out I can't review this PR because I'm the original author, but I'm happy for you to treat this comment as my approval @andrew-moores and approve it yourself.

andrew-moores (approved):

PR #1080, opened 2023-10-24
Olh 1122 emit an audit event when a user visits the triage page

andrew-moores:
## Proposed changes

Send an Audit Event via SQS to TxMA.  The audit event is emitted whenever someone visits the contact page.  

### What changed

The controller now calls out to a new EventService which sends messages via SQS.  The controller has been refactored and had some clarifying functions added.

Testing the controller has been done by creating a stub for the SQSClient.  This means the controller test knows too much about the way the EventService works but we will address this when we review the over all design with @alex9smith as a number of other areas where we can improve have been identified.

### Why did it change

To support the requirement to record an audit event with TxMA.  The interface to TxMA is via SQS.

## Testing

This has been tested in the DEV environment.  The logs from the ECS service show when the message is sent to TxMA.  The SQS queue has been reviewed and the expected message was seen.

### DIFF
--- a/src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
+++ b/src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
@@ -9,28 +9,158 @@ import {
   supportWebchatContact,
 } from "../../config";
 import { generateReferenceCode } from "../../utils/referenceCode";
+import { eventService } from "./event-service";
+import { AuditEvent, Extensions, Platform, User } from "./types";
 
 const CONTACT_ONE_LOGIN_TEMPLATE = "contact-govuk-one-login/index.njk";
+const MISSING_SESSION_VALUE_SPECIAL_CASE : string = "";
 
 export function contactGet(req: Request, res: Response): void {
-  const isAuthenticated = req.session?.user?.isAuthenticated;
+  updateSessionFromQueryParams(req.session, req.query);
+  const audit_event = buildAuditEvent(req);
+  logUserVisitsContactPage(audit_event);
+  sendUserVisitsContactPageAuditEvent(audit_event);
+  render(req, res);
+}
+
+const updateSessionFromQueryParams = (session: any, queryParams: any): void => {
+  if (isValidUrl(queryParams.fromURL as string)) {
+    session.fromURL = queryParams.fromURL;
+  } else {
+    logger.error("fromURL in request query for contact-govuk-one-login page did not pass validation");
+  }
+
+  copySafeQueryParamToSession(session, queryParams, 'theme');
+  copySafeQueryParamToSession(session, queryParams, 'appSessionId');
+  copySafeQueryParamToSession(session, queryParams, 'appErrorCode');
+
+  if (!session.referenceCode) {
+    session.referenceCode = generateReferenceCode();
+  }
+}
+
+const copySafeQueryParamToSession = (session: any, queryParams: any, paramName: string) => {
+  if (queryParams[paramName] && isSafeString(queryParams[paramName] as string)) {
+    session[paramName] = queryParams[paramName];
+  } else {
+    logger.error(`${paramName} in request query for contact-govuk-one-login page did not pass validation`);
+  }
+};
+
+const buildAuditEvent = (req: Request): AuditEvent => {
+  const session: any = req.session;
+  let sessionId: string;
+
+  if (userHasSignedIntoHomeRelyingParty(session)) {
+    sessionId = session.user.sessionId;
+  } else {
+    sessionId = MISSING_SESSION_VALUE_SPECIAL_CASE;
+  }
+
+  let persistentSessionId: string;
+
+  if (session.user?.persistentSessionId) {
+    persistentSessionId = session.user.persistentSessionId;
+  } else {
+    persistentSessionId = MISSING_SESSION_VALUE_SPECIAL_CASE;
+  }
+
+  const user: User = {
+    session_id: sessionId,
+    persistent_session_id: persistentSessionId,
+  };
+
+  const platform: Platform = {
+    user_agent: req.headers["user-agent"],
+  };
+
+  let appSessionId: string;
+
+  if (userHasComeFromTheApp(session)) {
+    appSessionId = req.session.appSessionId;
+  } else {
+    appSessionId = MISSING_SESSION_VALUE_SPECIAL_CASE;
+  }
+
+  const extensions: Extensions = {
+    from_url: req.session.fromURL,
+    app_error_code: req.session.appErrorCode,
+    app_session_id: appSessionId,
+    reference_code: req.session.referenceCode,
+  };
+
+  return {
+    timestamp: req.session.timestamp,
+    event_name: "HOME_TRIAGE_PAGE_VISIT",
+    component_id: "HOME",
+    user: user,
+    platform: platform,
+    extensions: extensions,
+  };
+}
+
+const userHasSignedIntoHomeRelyingParty = (session: any): boolean => {
+  return !!session.user?.sessionId;
+}
+
+const userHasComeFromTheApp = (session: any): boolean => {
+  return !!session.appSessionId;
+}
+
+const buildContactEmailServiceUrl = (req: Request): URL => {
+  const contactEmailServiceUrl: URL = new URL(getContactEmailServiceUrl());
+
+  if (req.session.fromURL) {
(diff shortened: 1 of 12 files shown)


src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -9,114 +9,182 @@ import {
...
+    event_name: "HOME_TRIAGE_PAGE_VISIT",
+    component_id: "HOME",
+    user: user,
+    platform: platform,
+    extensions: extensions,
+  };
+}
+
+const userHasSignedIntoHomeRelyingParty = (session: any): boolean => {
+  if (session.user?.sessionId) {
```
aenciso, line 105: Optional but this can be simplified to
`return !!session.user?.sessionId;`

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -9,114 +9,182 @@ import {
...
+    return true;
+  } else {
+    return false;
   }
-  if (theme) {
-    contactEmailServiceUrl.searchParams.append("theme", theme);
+}
+
+const userHasComeFromTheApp = (session: any): boolean => {
+  if (session.appSessionId) {
```
aenciso, line 113: and/or 
` return !!session.appSessionId;`

src/utils/sqs.ts
```diff
@@ -0,0 +1,32 @@
...
+import {
+  SendMessageCommand, SendMessageCommandOutput,
+  SendMessageRequest,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+import { logger } from "./logger";
+
+export function sqsService(): SqsService {
+  const send = async function (messageBody: string): Promise<any> {
+    const { AWS_REGION } = process.env;
```
aenciso, line 11: I think you can move the env variables and client outside the send function as they don't change between calls so there is no need to get them every time.

src/utils/sqs.ts
```diff
@@ -0,0 +1,32 @@
...
+    const { AUDIT_QUEUE_URL } = process.env;
+    const client = new SQSClient({ region: AWS_REGION });
+    const message: SendMessageRequest = {
+      QueueUrl: AUDIT_QUEUE_URL,
+      MessageBody: messageBody,
+    };
+
+    if (AUDIT_QUEUE_URL != null) {
```
aenciso, line 19: If you exit early we can reduce the nesting e.g.
`if (!AUDIT_QUEUE_URL) {
      logger.error(`Environment missing value for AUDIT_QUEUE_URL, cannot send ${messageBody}.`);
      return;
    }`

src/utils/sqs.ts
andrew-moores: Will this make any difference as the sqsService function is always called from the controller when sending an event?

src/utils/sqs.ts
```diff
@@ -0,0 +1,32 @@
...
+      try {
+        const result: SendMessageCommandOutput = await client.send(new SendMessageCommand(message));
+        logger.info(`Event sent with message id ${result.MessageId}`);
+      } catch (err) {
+        logger.error(`Failed to send message ${message.MessageBody} to SQS: ${err}`)
```
aenciso, line 24: Weird how sonarqube and `npm test:unit` report line 24 as uncovered despite the test being there.

src/utils/sqs.ts
aenciso: Not necessarily at this stage I think. Just thinking if we want to add more functions to the service would be nice to pull env variables at the top for clarity

andrew-moores: I tried this and it doesn't work because the sqsService function just returns the send function, it doesn't return an object with state and a send method.

src/utils/sqs.ts
andrew-moores: @alex9smith is this something you've seen before?

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -9,114 +9,174 @@ import {
...
-    contactPhoneEnabled: supportPhoneContact(),
-    showContactGuidance: showContactGuidance(),
-    showSignOut: isAuthenticated && !isLoggedOut,
-    referenceCode,
-    contactEmailServiceUrl: contactEmailServiceUrl,
-    webchatSource: getWebchatUrl(),
-    currentUrl: originalUrl,
-    baseUrl,
-    language,
+  if (userHasSignedIntoHomeRelyingParty(session)) {
```
aenciso, line 56: This can be simplified to
`const sessionId = session.user?.sessionId ?? MISSING_SESSION_VALUE_SPECIAL_CASE;`

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -9,114 +9,174 @@ import {
...
+    sessionId = session.user.sessionId;
+  } else {
+    sessionId = MISSING_SESSION_VALUE_SPECIAL_CASE;
+  }
+
+  let persistentSessionId: string;
+
+  if (session.user?.persistentSessionId) {
```
aenciso, line 64: Similarly, this one to
`const persistentSessionId = session.user?.persistentSessionId ?? MISSING_SESSION_VALUE_SPECIAL_CASE;`

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
andrew-moores: I want to keep the clarifying function calls as they provide information about what the check actually means to the user.

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
andrew-moores: Similar to previous comment we can update this to clarify what the purpose behind the check is.

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -9,114 +9,174 @@ import {
...
 
 const CONTACT_ONE_LOGIN_TEMPLATE = "contact-govuk-one-login/index.njk";
+const MISSING_SESSION_VALUE_SPECIAL_CASE : string = "";
 
-export function contactGet(req: Request, res: Response): void {
-  const isAuthenticated = req.session?.user?.isAuthenticated;
-  let isLoggedOut = req.cookies?.lo;
-  if (typeof isLoggedOut === "string") {
-    isLoggedOut = JSON.parse(isLoggedOut);
+export function contactGet(req: Request, res: Response, ): void {
```
aenciso, line 18: extra comma

src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
```diff
@@ -9,114 +9,174 @@ import {
...
+  } else {
+    sessionId = MISSING_SESSION_VALUE_SPECIAL_CASE;
+  }
+
+  let persistentSessionId: string;
+
+  if (session.user?.persistentSessionId) {
+    persistentSessionId = session.user.persistentSessionId;
+  } else {
+    persistentSessionId = MISSING_SESSION_VALUE_SPECIAL_CASE
```
aenciso, line 67: Missing semi-colon

src/utils/sqs.ts
```diff
@@ -0,0 +1,33 @@
+import { SqsService } from "./types";
+import {
+  SendMessageCommand, SendMessageCommandOutput,
+  SendMessageRequest,
+  SQSClient,
+} from "@aws-sdk/client-sqs";
+import { logger } from "./logger";
+
+export function sqsService(): SqsService {
```
aenciso, line 9: I think we could probably follow the same pattern as all the other services? Have a look at snsService, basically we abstract all the aws config to aws.ts

src/utils/sqs.ts
alex9smith: It's not but I don't 100% trust Sonar's code coverage metrics - we've had a few instances where it's complained there's no tests for a file when there clearly are. I'm not sure about the lcov coverage though, I'll try and investigate

andrew-moores: Won't stop approval of this PR.

aenciso (approved):

PR #1090, opened 2023-10-30
OLH-1122 - add missing timestamp to audit event.

andrew-moores:
## Proposed changes

Add timestamp to audit event.

### What changed
The audit event being sent to TxMA now creates a timestamp rather than trying to retrieve one from the session.

### Why did it change
The timestamp is a mandatory field that causes TxMA to send event to the DLQ if missing.  The session did not contain a timestamp so we have changed to generating one at the time the event is created.

## Testing

I tested in DEV.  I could see the new audit event had a timestamp.  I did not see the event on the DLQ.


alex9smith (approved):
I've also checked the events going to SQS in dev and verified they have timestamps now.

aenciso (approved):

PR #1092, opened 2023-11-01
OLH-1169: launch webchat from link on triage page

danacotoran:
### What changed
JIRA for more detail: https://govukverify.atlassian.net/browse/OLH-1169

[OLH-1169: Launch webchat from "Use webchat" link](https://github.com/alphagov/di-account-management-frontend/commit/f0828e5fb304c15c8d8b5779c84b5c18862159ba) 

Enable launching the webchat from the "Use webchat" link on the page.
Even though the element looks like a link, the markup for `<button>` has been used in order to make it semantically correct. 
The reason it should be a <button> instead of an <a> even though it looks link like is that its purpose is not navigational – it has no `href`

https://github.com/alphagov/di-account-management-frontend/assets/7116819/45659b5b-2be9-44e0-8153-b33f69ad48ec

[OLH-1169: \</br> should be \<br>](https://github.com/alphagov/di-account-management-frontend/commit/2c08344d93a857a34d71ec5db0d49bcfa6222a61) 

Because that is more correct. `</br>` is actually not technically valid, though browsers will accept it and render correctly anyway. 

### Why did it change

The webchat widget automatically creates a small purple circle on the user’s screen when it loads, which is a trigger for launching it. This behaviour can’t be prevented.

We want to augment this by allowing a user to launch the webchat from the page content, which will be more obvious. This it to make it easier for users to access the webchat widget when it goes live.

### Related links
N/A

## How to review
Check that it works as outlined in the JIRA ticket.
Note: there is a second part to this work, which is outlined in [this comment](https://govukverify.atlassian.net/browse/OLH-1169?focusedCommentId=100803) on JIRA: ensuring that "Use webchat" is conditional upon presence of JS as it currently displays whether JS is enabled in the browser or not. We will be recording that as a separate ticket

### DIFF
--- a/src/components/contact-govuk-one-login/index.njk
+++ b/src/components/contact-govuk-one-login/index.njk
@@ -105,8 +105,9 @@
         <h3 class="govuk-heading-m">{{'pages.contact.section3.webchat.heading' | translate}}</h3>
 
         {% for paragraph in webchatParagraphs %}
-          <p class="govuk-body">{{ paragraph | safe | replace('[useWebchatLinkHref]', 'pages.contact.section3.webchat.useWebchatLinkHref' | translate ) }}</p>
+          <p class="govuk-body">{{ paragraph | safe  }}</p>
         {% endfor %}
+        <p class="govuk-body"><button type="button" class="launch-webchat-link" data-launch-webchat>{{ 'pages.contact.section3.webchat.linkText' | translate }}</button></p>
       </section>
     {% endif %}
 
@@ -133,6 +134,14 @@
   <script id="smartagent" type="module" defer 
     src="{{ webchatSource }}" 
     data-company="hgsgds" data-brand="hgsgds">
+  </script>
+  <script type="text/javascript" nonce='{{scriptNonce}}'>
+    var launchWebchatButton = document.querySelector("[data-launch-webchat]");
+    launchWebchatButton.addEventListener("click", function() {
+      if (window._sa) {
+        window._sa.openChat();
+      }
+    })
   </script>
 {% endif %}
 {% endblock %}
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -333,3 +333,16 @@ $nav-active-border-thickness: 5px;
   color: $govuk-border-colour;
 }
 
+.launch-webchat-link {
+  margin: 0;
+  padding: 0;
+  border: 0;
+  background: none;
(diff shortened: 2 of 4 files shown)


src/components/contact-govuk-one-login/index.njk
```diff
@@ -134,5 +135,14 @@
     src="{{ webchatSource }}" 
     data-company="hgsgds" data-brand="hgsgds">
   </script>
+  <script type="text/javascript" nonce='{{scriptNonce}}'>
+    var launchWebchatButton = document.querySelector("[data-launch-webchat]")
```
aenciso, line 139: I'm not entirely sure why our eslint rules are not being applied on save but I think we should have semicolons at the end of this line and the ones inside the if.

src/components/contact-govuk-one-login/index.njk
```diff
@@ -134,5 +135,14 @@
     src="{{ webchatSource }}" 
     data-company="hgsgds" data-brand="hgsgds">
   </script>
+  <script type="text/javascript" nonce='{{scriptNonce}}'>
+    var launchWebchatButton = document.querySelector("[data-launch-webchat]");
+    launchWebchatButton.addEventListener("click", function(e) {
+      if (window._sa) {
+        e.preventDefault();
```
aenciso, line 142: I think we don't need to call preventDefault as the button doesn't do anything. Since the button has a type of "button", it doesn't have a default action associated with it when clicked, like submitting a form or navigating to a different page. Therefore, calling e.preventDefault() doesn't have any effect in this context.

danacotoran: That's true. It's because I tried implementing this with a link element initially, before I did some research and remembered that `<a>` was not the correct tag to use in this particular instance. Good catch.

src/components/contact-govuk-one-login/index.njk
```diff
@@ -134,5 +135,13 @@
     src="{{ webchatSource }}" 
     data-company="hgsgds" data-brand="hgsgds">
   </script>
+  <script type="text/javascript" nonce='{{scriptNonce}}'>
+    var launchWebchatButton = document.querySelector("[data-launch-webchat]");
+    launchWebchatButton.addEventListener("click", function(e) {
```
aenciso, line 140: Sorry one more, we don't need to pass e

aenciso (approved):

PR #1093, opened 2023-11-01
OLH-1184 - Try strict-dynamic directive for the webchat

alex9smith:
## Proposed changes

[OLH-1184] - Try strict-dynamic directive for the web chat 
Change our  Content Security Policy (CSP) to **_strict dynamic_** to prevent cross site scripting. 
This change will NOT support older browser versions that does not support strict dynamic. As that will mean the site is still susceptible to CSS attack with older versions.
JIRA Ticket: https://govukverify.atlassian.net/browse/OLH-1184

### What changed
- Make nonce available to frontend
- Include a nonce in the <script> tags that load the initial webchat script
- Switch the webchat from using the ‘unsafe-inline' Content Security Policy (CSP) directive for scripts to 'strict-dynamic'.
- Test to ensure all required scripts are still loading  even with **_strict dynamic_** configured

### Why did it change
Using 'unsafe-inline' for scripts in the CSP is an unacceptable security risk

### Related links
Configuration Doc: https://content-security-policy.com/strict-dynamic/

## Checklists
Test in higher environment to ensure DI- one login home is still able to load all existing scripts and pages function correctly

Please see below sample of the CSP in the dev enironment, was able to use webchat, send messsages and closes fine, No CSP related error in the console

`default-src 'self';style-src 'self' https://*.smartagent.app 'unsafe-inline';script-src 'self' 'nonce-0d423d3e0184529a3fad04c4e8ae34b7' 'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=' https://www.googletagmanager.com https://www.google-analytics.com https://ssl.google-analytics.com https://*.smartagent.app 'strict-dynamic';script-src-attr 'self' 'unsafe-inline';img-src 'self' data: https://www.googletagmanager.com https://www.google-analytics.com https://*.s3.eu-west-2.amazonaws.com;object-src 'none';connect-src 'self' https://www.google-analytics.com https://*.smartagent.app;form-action 'self' https://*.account.gov.uk;media-src 'self' https://*.s3.eu-west-2.amazonaws.com;frame-src 'self' https://*.smartagent.app;base-uri 'self';font-src 'self' https: data:;frame-ancestors 'self';upgrade-insecure-requests`

N/A

## Testing

- Open: https://home.dev.account.gov.uk/contact-gov-uk-one-login
- Open Developer Console on the page
- Check console and ensure no error exist with loading required javascript e.g. application.js, all.js.
- Verify you can initiate web chat
- Verify strict domain by checking the response header for the CSP. It will be titled "content-security-policy

## How to review

[OLH-1184]: https://govukverify.atlassian.net/browse/OLH-1184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/common/layout/base.njk
+++ b/src/components/common/layout/base.njk
@@ -113,8 +113,8 @@
 
 {% block bodyEnd %}
     {% block scripts %}{% endblock %}
-    <script type="text/javascript" src="/public/scripts/application.js"></script>
-    <script type="text/javascript" src="/public/scripts/all.js"></script>
+    <script type="text/javascript" src="/public/scripts/application.js" nonce='{{scriptNonce}}'></script>
+    <script type="text/javascript" src="/public/scripts/all.js" nonce='{{scriptNonce}}'></script>
     <script type="text/javascript" nonce='{{scriptNonce}}'>
         window
             .GOVSignIn
--- a/src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
+++ b/src/components/contact-govuk-one-login/contact-govuk-one-login-controller.ts
@@ -60,6 +60,7 @@ const render = (req: Request, res: Response): void => {
     currentUrl: originalUrl,
     baseUrl,
     language,
+    nonce: res.locals.scriptNonce,
   };
 
   res.render(CONTACT_ONE_LOGIN_TEMPLATE, data);
--- a/src/components/contact-govuk-one-login/index.njk
+++ b/src/components/contact-govuk-one-login/index.njk
@@ -133,7 +133,8 @@
 {% if contactWebchatEnabled %}
   <script id="smartagent" type="module" defer 
     src="{{ webchatSource }}" 
-    data-company="hgsgds" data-brand="hgsgds">
+    data-company="hgsgds" data-brand="hgsgds"
+    nonce="{{ nonce }}">
   </script>
   <script type="text/javascript" nonce='{{scriptNonce}}'>
     var launchWebchatButton = document.querySelector("[data-launch-webchat]");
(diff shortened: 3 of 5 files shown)


deploy/template.yaml
```diff
@@ -143,7 +143,7 @@ Mappings:
       LOGSLEVEL: "debug"
       SUPPORTTRIAGEPAGE: "1"
       WEBCHATSOURCEURL: "https://uat-chat-loader.smartagent.app/loader/main.js"
-      SUPPORTWEBCHATCONTACT: "0"
+      SUPPORTWEBCHATCONTACT: "1"
```
andrew-moores, line 146: Are we leaving webchat on in DEV going forward?  It won't be prod like if we do.

andrew-moores (approved):

alex9smith:
@peterfajemisincabinetoffice could you squash / fixup these commits please? It's fine to have multiple commits in a PR but we shouldn't have the merge commits from main or the merge conflicts as their own commits as it leaves us with a messy history in `main`. 

I think here it'll be easiest to pull the latest version of `main`, rebase this whole branch onto it then squash all the commits into a single one.

deploy/template.yaml
peterfajemisincabinetoffice: Hi @alex9smith , as per the questions above from @andrew-moores , would like your input here pls?

alex9smith: I think we should switch this back off in dev - we know we're not going to be working on the webchat for at least this sprint. Dev isn't going to be identical to production all the time but we should try and keep it as close as possible for longer-term things.

PR #1094, opened 2023-11-01
OLH-1173 - Emit an audit event when a user clicks the email link on the triage page

aenciso:
## Proposed changes
[OLH-1173] Emit an audit event when a user clicks the email link on the triage page

### What changed

Added TypeScript definitions specifically for our Session and User types.

Lots of files changed due to TS being added mostly.

Create a new route in app.config.ts  register it in the contact router, and add a controller function for it.

Move the logic that appends the URL parameters to the current link across to the new controller

Update the new controller to send the audit event

Update the link on the triage page to point to our new route

### Why did it change

We need a way of knowing when a user clicks on this link that doesn’t involve any client-side Javascript, as users with JS disabled or on old devices will be pushed towards this as a contact channel because they won’t be able to use the webchat. 

## How to review

[OLH-1173]: https://govukverify.atlassian.net/browse/OLH-1173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


@types/express/index.d.ts
```diff
@@ -1,14 +1,35 @@
...
+declare module "express-session" {
+  interface Session {
+    timestamp?: number;
+    nonce?: string;
+    state?: string;
+    currentURL?: string;
+    user?: User;
+    user_id?: string;
+    referenceCode?: string;
+    queryParameters: QueryParameters;
```
andrew-moores, line 14: Why is this the only optional parameter in the session?

andrew-moores, line 14: Why is this the only optional parameter in the session?

.env.sample
```diff
@@ -26,4 +26,5 @@ SUPPORT_WEBCHAT_CONTACT=1
 SUPPORT_PHONE_CONTACT=1
 SHOW_CONTACT_GUIDANCE=1
 CONTACT_EMAIL_SERVICE_URL=https://signin.staging.account.gov.uk/contact-us-from-triage-page
-AUDIT_QUEUE_URL=
+AUDIT_QUEUE_URL=http://localhost:4566/000000000000/localstack-queue
+AWS_REGION=us-east-1
```
andrew-moores, line 30: Why is the being set?  The provision.sh script does all its work in eu-west-2 (which we don't really need for localstack) which is at odds with this setting.

.env.sample
```diff
@@ -26,4 +26,5 @@ SUPPORT_WEBCHAT_CONTACT=1
 SUPPORT_PHONE_CONTACT=1
 SHOW_CONTACT_GUIDANCE=1
 CONTACT_EMAIL_SERVICE_URL=https://signin.staging.account.gov.uk/contact-us-from-triage-page
-AUDIT_QUEUE_URL=
+AUDIT_QUEUE_URL=http://localhost:4566/000000000000/localstack-queue
```
andrew-moores, line 29: The queue name created in provision.sh is `audit-events`.  I think the url without `000000...` will work as well.

aenciso: Forgot to update after merging your changes. Deleted now.

.env.sample
aenciso: Forgot to update after merging your changes. Deleted now.

@types/express/index.d.ts
aenciso: Actually, all are optional but queryParameters. It has been updated now.

PR #1096, opened 2023-11-03
OLH-1123 - Return a 403 error when a user fails to sign in

peterfajemisincabinetoffice:
## Proposed changes
When an access denied error is sent from auth-frontend, rather than handling the error, we let it propagate which triggers another error and causes a 500 error to be thrown rather than the workflow for handling access denied errors.
JIRA Ticket: https://govukverify.atlassian.net/browse/OLH-1123

### What changed
In auth-callback controller, we check the query params for access denied error, if founds, we will then update the response status to 403.  This triggers the internal-error-handler and redirects to the configured session expired page.

Also this will stop the triggering of the [CloudWatch Alarm | account-mgmt-frontend-ApiGateway5xxStatusAlarm | eu-west-2 | Account: 026991849909](https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#alarm:name=account-mgmt-frontend-ApiGateway5xxStatusAlarm) in https://gds.slack.com/archives/C04D3SQNJ4B

### Why did it change

So that the Home team handles access denied errors correctly and not throw error 500 incorrectly.

### Related links
https://govukverify.atlassian.net/browse/OLH-1123
https://gds.slack.com/archives/C04D3SQNJ4B

N/A

## Testing
- Retrieved the access denied logs in production and examined the request/response and stack trace
- In Local environment, trigger the same request as per the prod logs in local environment and dev environment
- Observe the behaviour is as per expected

## How to review
Code changes are minimal and self descriptive in the PR
To test, please use: http://home.dev.account.gov.uk/auth/callback?error=access_denied&state=m0H_2VvrhKR0qA

### DIFF
--- a/src/components/oidc-callback/call-back-controller.ts
+++ b/src/components/oidc-callback/call-back-controller.ts
@@ -1,6 +1,6 @@
 import { Request, Response } from "express";
 import { CallbackParamsType, TokenSet, UserinfoResponse } from "openid-client";
-import { PATH_DATA, VECTORS_OF_TRUST } from "../../app.constants";
+import { HTTP_STATUS_CODES, OIDC_ERRORS, PATH_DATA, VECTORS_OF_TRUST } from "../../app.constants";
 import { ExpressRouteFunc } from "../../types";
 import { ClientAssertionServiceInterface } from "../../utils/types";
 import { clientAssertionGenerator } from "../../utils/oidc";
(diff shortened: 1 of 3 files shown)


src/components/oidc-callback/call-back-controller.ts
```diff
@@ -47,6 +47,9 @@ export function oidcAuthCallbackGet(
 ): ExpressRouteFunc {
   return async function (req: Request, res: Response) {
     const queryParams: CallbackParamsType = req.oidc.callbackParams(req);
+    if(queryParams?.error === OIDC_ERRORS.ACCESS_DENIED) {
```
andrew-moores, line 50: Missing space (that prettier doesn't catch?)
```suggestion
    if (queryParams?.error === OIDC_ERRORS.ACCESS_DENIED) {
```

andrew-moores (approved):

PR #1102, opened 2023-11-09
BAU: Enable webchat in integration

alex9smith:
### What changed

Enable webchat in integration
### Why did it change

We thought our suppliers had access to the GDS VPN and therefore could access the webchat in our IP-restricted staging environment. That's turned out not to be true.

We could gather their IP ranges and add them to the allowlist, but it's easier to simply enable the webchat in integration where we don't restrict by IP address.

## Testing

N/A


aenciso (approved):

andrew-moores:
Its also on in Staging, is that intended?

andrew-moores (approved):

alex9smith:
I think that's fine for now. We can get this merged & deployed then clean up the environments if we think we need to

PR #1103, opened 2023-11-09
OLH-1219: Remove LITE from Services you can use list

peterfajemisincabinetoffice:
## Proposed changes
Remove LITE from the services you can use list on the Your services page

### What changed
Removed the link to the LITE service from the translations object for English and Welsh

The LITE link on this list links into the authenticated part of their service which means a user who clicks on this link gets signed in to LITE and gets a service card for them, which may be accidental.

LITE are still in private beta so don’t have a start page we can link to from this list. Their being in private beta also means we probably shouldn’t be linking to them from this list anyway.

### Related links
[

](https://govukverify.atlassian.net/browse/OLH-1219)

## Testing

Navigate to 

1. https://home.staging.account.gov.uk//your-services?lng=cy
2. https://home.staging.account.gov.uk//your-services?lng=en

In both pages, you should not see the LITE service listed any longer

### DIFF
--- a/src/locales/cy/translation.json
+++ b/src/locales/cy/translation.json
@@ -241,10 +241,6 @@
               "href": "https://www.socialworkengland.org.uk/registration/apply-for-registration"
             },
             {
-              "text": "LITE (Trwyddedu ar gyfer Masnach a Menter Ryngwladol)",
-              "href": "https://exporter.lite.private-beta.service.trade.gov.uk/"
-            },
-            {
               "text": "Llofnodwch eich gweithred morgais",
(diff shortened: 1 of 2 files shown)


andrew-moores:
Is this all that is required or should we be removing the LITE client entries from the clientRegistry as well?

peterfajemisincabinetoffice:
> Is this all that is required or should we be removing the LITE client entries from the clientRegistry as well?

Thank Andrew very good question and session with Ben, on the basis of the response we got, we would not need to remove the LITE client entries for now.

andrew-moores (approved):

PR #1125, opened 2023-12-08
OLH-1279 Investigate reference codes with multiple session ids.

andrew-moores:
## Proposed changes

We can't see anything obvious in the logs that is causing this scenario but we can see that there are errors being logged when the session is being destroyed.  Its possible that this is related to this issue.  However, its not possible to attribute one of these errors to a specific user journey as we don't have transaction tracing in our logs so this change is to add a trace element to our logs that will contain the session id to facilitate this.

### What changed

Updating logging messages to include a trace id to link a users log messages easily.  Using session id for the value of the trace. Tidied the app.constants ERROR_MESSAGES constant to remove unused messages. Introduced new constant to app.constants for info log messages. Added additional logging to logout-controller to see if the failure to destroy the session is causing the issue with the reference code persisting between sessions.

### Why did it change

To support transaction tracing and some tidy up of the text used for the logging messages.

### Related links

Also see https://github.com/govuk-one-login/team-manual/pull/389 update to our section of the DI Manual.

## Testing

Unit and integration tests are passing but we will need to ship logs to Splunk to confirm it works as expected.


alex9smith (approved):
Looks good to me 👍🏻

PR #1130, opened 2023-12-13
Olh 1279 investigate reference codes with multiple session ids

andrew-moores:
## Proposed changes

Add checks before reusing a reference code to ensure we always create a new reference code when the users gs session id changes.

### What changed

The `update-session-middleware` module was changed to include more checks.  The gs session id is now added to the express session along with the reference code.

### Why did it change

A reference code should be unique to a gs session but the current implementation allowed reference codes to persist between gs sessions and to persist from a scenario where a user did not have a gs session.

## Testing

I have added unit tests and updated existng tests.  To test the fix in the dev environment you will need to manually add the gs cookie and manipulate it using the browsers developer tools.


aenciso (approved):

src/app.ts
```diff
@@ -71,13 +71,21 @@ async function createApp(): Promise<express.Application> {
...
   app.use(express.json());
   app.use(express.urlencoded({ extended: false }));
 
   app.use(cookieParser());
+  app.use(setLocalVarsMiddleware);
+  app.use(loggerMiddleware);
+
+  app.use((req, res, next) => {
+    req.log = req.log.child({
+      trace: "res.locals.trace",
```
alex9smith, line 85: Is this meant to be a string instead of the variable?

src/components/track-and-redirect/track-and-redirect-controller.ts
```diff
@@ -25,7 +28,10 @@ export const buildContactEmailServiceUrl = (req: Request): URL => {
         ]
       );
     } else {
-      logger.info(`Missing ${paramValue} in the request or session.`);
+      logger.info(
+        { trace: res.locals.sessionId },
```
alex9smith, line 32: Why are we using the session ID as the trace here instead of the `res.locals.trace`?

src/utils/yourServices.ts
```diff
@@ -61,7 +65,7 @@ export const presentYourServices = async (
 export const getAllowedListServices = async (
   subjectId: string
 ): Promise<Service[]> => {
-  const userServices = await getServices(subjectId);
+  const userServices = await getServices(subjectId, "session-id");
```
alex9smith, line 68: Should this be
```suggestion
  const userServices = await getServices(subjectId, trace);
```

alex9smith (commented):
The logic all looks good to me. I've left a couple of questions about some of the trace values

src/app.ts
andrew-moores: Changed to reference the variable.

src/components/track-and-redirect/track-and-redirect-controller.ts
andrew-moores: Changed to use trace from res.locals.

src/utils/yourServices.ts
andrew-moores: Changed to use trace from res.locals.

alex9smith (approved):

PR #1137, opened 2023-12-18
OLH-1292 Onboard AAS.

andrew-moores:
## Proposed changes

Add AAS to available services.

### What changed

Config updated.

### Why did it change

To allow the new service to be accessed from the services list and to present a service card when the service has been visited previously.


peterfajemisincabinetoffice (approved):
LGTM

PR #1139, opened 2023-12-20
OLH-1224 Sign out link as POST request

aenciso:
## Proposed changes
Change sign-out link from GET to POST request

### What changed

There were two sign-out links, one inside the header and another one on the navigation. Both links were changed to buttons

### Why did it change

The use of POST for a logout feature is strongly recommended to enhance security, adhere to web standards and HTTP specifications, and ensure a better user experience. It mitigates the risk of CSRF attacks, keeps user actions intentional, and aligns with the principles of safe web navigation.

## Testing

Tested locally for both signed in and not signed in users to make sure the link only appears on logged in users only.
/sign-out path does not work any more unless it's a valid POST request

## How to review

Replicate the testing above in different environments

### DIFF
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -67,7 +67,7 @@ $govuk-new-link-styles: true;
 
 .govuk-summary-list--activity-log {
   .govuk-summary-list__key {
-    @include govuk-font($size: 24, $weight: bold)
+    @include govuk-font($size: 24, $weight: bold);
   }
 
   @include govuk-media-query($from: "tablet") {
@@ -201,16 +201,89 @@ $govuk-new-link-styles: true;
 $nav-link-padding: govuk-spacing(3);
 $nav-link-outline-thickness: 2px;
 $nav-active-border-thickness: 5px;
+$nav-link-padding-left: $nav-link-padding - 2;
+$nav-button-padding-bottom: $nav-link-padding + $nav-link-outline-thickness;
 
+/* Combined styles for buttons */
+.account-header-navigation__button, .account-navigation__button {
+  border: 0 none;
+  cursor: pointer;
+  background-color: transparent;
+  padding: 0;
+  margin: 0;
+
+  &:focus {
+    outline: none;
+    span {
+      @include govuk-focused-text;
+    }
+  }
+}
+
+/* Specific styles for .account-header-navigation__button */
+.account-header-navigation__button-content {
+  @include govuk-font($size: 16);
+  color: govuk-colour("white");
+  @include govuk-link-style-inverse;
+
+  &:not(:hover) {
+    text-decoration: none;
+  }
+
+  &:hover {
+    text-decoration: underline;
+    text-decoration-thickness: 3px;
+
+    @if $govuk-link-underline-offset {
+      text-underline-offset: $govuk-link-underline-offset;
+    }
+  }
+
+  &:focus {
+    @include govuk-focused-text;
+  }
+}
+
+// /* Positioning for the navigation form */
+
+.account-header-navigation__form {
+  position: absolute;
+  padding: 5px;
+  right: 0;
+  bottom: 7px;
+  @include govuk-media-query($until: mobile) {
+    position: relative;
+    float: left;
+    padding-left: 0;
+  }
+}
+
+// /* Specific styles for .account-navigation__button */
+.account-navigation__button {
+  color: $govuk-link-colour;
+  font-size: inherit;
+
+  &:focus {
+    span {
+      padding: $nav-link-padding-left 0 16px;
+    }
+  }
+}
+
+.account-navigation__button-content,
 .account-navigation__link {
   @extend .govuk-link;
   display: inline-block;
   font-weight: bold;
-  padding: $nav-link-padding - $nav-link-outline-thickness 0 $nav-link-padding;
+  padding: $nav-link-padding-left 0 govuk-spacing(2);
   margin-top: $nav-link-outline-thickness;
 
   &:not(:hover) {
     text-decoration: none;
+  }
+
+  &:focus {
+    padding: $nav-link-padding-left 0 $nav-link-padding;
   }
 
   &:visited {
@@ -226,18 +299,22 @@ $nav-active-border-thickness: 5px;
   }
 
   .account-navigation__list-item--active & {
-    padding: $nav-link-padding - $nav-link-outline-thickness 0 $nav-link-padding - $nav-active-border-thickness;
     border-bottom: $nav-active-border-thickness solid $govuk-link-colour;
 
     &:focus {
       border-bottom: none;
-      padding: $nav-link-padding - $nav-link-outline-thickness 0 $nav-link-padding;
+      padding: $nav-link-padding-left 0 $nav-link-padding;
+      margin-top: $nav-link-outline-thickness;
     }
 
     &:hover {
       border-color: $govuk-link-hover-colour;
     }
   }
+}
+
+.account-navigation__button-content {
+  padding: $nav-link-padding-left 0 $nav-link-padding-left;
 }
 
 .your-services__heading {
@@ -353,15 +430,15 @@ $nav-active-border-thickness: 5px;
   }
 }
 
-.sa-chat-tab { 
-  // limit size of webchat button when zoom is applied 
-  max-width: 25vw !important; 
+.sa-chat-tab {
+  // limit size of webchat button when zoom is applied
+  max-width: 25vw !important;
   max-height: 25vw !important;
   bottom: 2vh !important;
   right: 2vw !important;
   left: auto !important;
   @include govuk-media-query($from: tablet) {
-    max-width: 20vw !important; 
+    max-width: 20vw !important;
     max-height: 20vw !important;
   }
 }
--- a/src/components/common/layout/header.njk
+++ b/src/components/common/layout/header.njk
@@ -1,0 +1,29 @@
+<header class="govuk-header {{ headerClasses }}" role="banner" data-module="govuk-header">
+    <div class="govuk-header__container govuk-width-container">
+        <div class="govuk-header__logo">
+            <a href="{{ accountHome }}" class="govuk-header__link govuk-header__link--homepage" style="margin-right: 20px;">
+                <span class="govuk-header__logotype">
+                  <!--[if gt IE 8]><!-->
+                  <svg aria-hidden="true" focusable="false" class="govuk-header__logotype-crown" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 132 97" height="30" width="36">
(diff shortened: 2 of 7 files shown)


src/components/common/layout/header.njk
```diff
@@ -0,0 +1,34 @@
...
+                    GOV.UK
+                  </span>
+                </span>
+                <span class="govuk-header__product-name">{{ 'general.header.productName' | translate }}</span>
+            </a>
+        </div>
+        {% if showSignOut %}
+            <div class="govuk-header__content">
+                <nav aria-label="Menu" class="govuk-header__navigation ">
+                    <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide menu" hidden="">Menu</button>
```
danacotoran, line 21: I was going to note that the "Menu" button is not translated into Welsh which might be an issue.

But upon giving it more thought: since we're no longer importing the component from the DS directly and we have finer control over its distinct elements, perhaps it would be a good idea to remove the "Menu" toggle completely seeing as it's usefulness is limited in this case. The "Sign out" link in the black bar is an edge case and there is no real reason to hide it on mobile.

It might be one to check with the UCD team first.

src/assets/scss/application.scss
```diff
@@ -239,6 +239,29 @@ $nav-active-border-thickness: 5px;
   }
 }
 
+.account-header-navigation__button {
+  @include govuk-font($size: 19, $weight: bold);
+  border: 0 none;
+  cursor: pointer;
```
danacotoran, line 245: This is the same selector and rules as L249 (minus font size). Since L249 and below takes precedence, can we remove this?

src/components/common/layout/navigation.njk
```diff
@@ -8,18 +8,21 @@
...
+    <div class="govuk-width-container">
+        <ul class="account-navigation__list">
+            {% for item in navigationItems %}
+                <li class="account-navigation__list-item {{ 'account-navigation__list-item--active' if item.active else '' }}">
+                    <a class="account-navigation__link" href="{{ item.href }}">{{ item.text }}</a></li>
+            {% endfor %}
+            <li class="account-navigation__list-item">
+                <form action="{{ accountSignOut }}" method="post">
+                    <input type="hidden" name="_csrf" value="{{csrfToken}}"/>
+                    <button class="govuk-link account-navigation__link {{ 'account-navigation__list-item--active' if item.active else '' }} account-navigation__button" type="submit">{{ 'general.header.signOutLink' | translate }}</button>
```
danacotoran, line 23: A couple questions:
What is the role of `govuk-link`?
The active modifier isn't working. That being said, is it even needed? Seeing as clicking Sign out signs out out automatically and redirects you

src/assets/scss/application.scss
```diff
@@ -239,6 +239,29 @@ $nav-active-border-thickness: 5px;
...
+
+.account-header-navigation__button {
+  @include govuk-font($size: 16, $weight: bold);
+  border: 0 none;
+  cursor: pointer;
+  color: govuk-colour(white);
+  padding: 0;
+}
+
+.account-navigation__button {
```
danacotoran, line 257: Some of the button reset styling is repeated between `.account-header-navigation__button` and `.account-navigation__button` so it could be possible to do
```
.account-header-navigation__button,
.account-navigation__button {
   ...stuff...
}
```

danacotoran (commented):
@aenciso thanks for this, just had a couple comments (I left the first comment accidentally when I meant to include it in the overall review, sorry!). 

It looked good in the browsers I've tested it in on Browserstack but I have been having issues with testing local on Safari. I'm not sure whether that's one you managed to test.

src/assets/scss/application.scss
aenciso: Of course, I don't know how I missed this, probably end of the year brain fog

src/components/common/layout/header.njk
aenciso: Agree, maybe we can discuss it during UCD huddle?

src/components/common/layout/navigation.njk
aenciso: Since the element is a button, the role is implicitly button. govuk-link is purely to style the button as link.
I have also removed the active modifier.

src/assets/scss/application.scss
aenciso: Done

src/components/common/layout/navigation.njk
danacotoran: Yes, I suppose what I mean is more... is `govuk-link` actually needed to style this item the same as the rest of the items in the navigation, or is the `account-navigation__link` class sufficient?

aenciso: You're right, it isn't needed.

src/components/common/layout/header.njk
aenciso: Actually I had a look at the HMRC website and they apply this approach, I can only see Sign out instead of the menu so I'll go ahead and remove it completely as it makes more sense.

danacotoran (commented):
This is looking much better now! I managed to spot a couple small issues though, sorry.

The alignment is a little off compared to the version that is currently live. 

Screenshot 2024-01-11 at 19 01 16

I also finally managed to test this on Safari (by deploying it to dev first) and I spotted that the "Sign out" in the black header doesn't have a hover state.

The one in the secondary nav does have a hover state, however the underline is thinner than the other links ❓ Strange one – it might be to do with the `text-decoration-thickness` property that is applied to the links as well as the button. This only seems to happen in more recent versions of Safari, from 16.5.

Another Safari-related issue is quite subtle, but it's showing up as some additional spacing under the secondary navigation (screenshots of live account vs dev version from Safari 14.1). The same issue can be observed on some fairly recent versions of iOS Safari

Screenshot 2024-01-11 at 19 43 43

Screenshot 2024-01-11 at 19 40 11

src/assets/scss/application.scss
```diff
@@ -239,6 +239,36 @@ $nav-active-border-thickness: 5px;
...
+  @include govuk-font($size: 16, $weight: bold);
+  color: govuk-colour(white);
+  padding: 0;
+  text-decoration: none;
+  position: absolute;
+  top: govuk-spacing(4);
+  right: 0;
+  margin: 0;
+  &:hover {
+    text-decoration: solid underline 3px;
```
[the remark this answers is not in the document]
aenciso: I couldn't use mixins for this particular bit because it says:
   // Avoid using the `govuk-link-common` mixin because the links in the header
    // get a special treatment, because:
    //
    // - underlines are only visible on hover
    // - all links get a 3px underline regardless of text size, as there are
    //   multiple grouped elements close to one another and having slightly
    //   different underline widths looks unbalanced

So I used standard govuk-frontend classes instead

danacotoran (commented):
Hi,
Looks better now! Unfortunately I still found some small issues. 

### One bit that is not really an issue 
I think it would be a good idea to have one of the UCD team sign off the visual placement of the Sign out link. We've removed the Menu toggle so on mobile it sits very differently from where it used to. 

It'd be best to chat with Anisa/Ale/Mel and confirm whether it looks okay as it's a visual change from what it was before. 

Screenshot 2024-01-18 at 12 02 20

### Other bits that actually kind of are issues

1. Hovering over the "Sign out" button in the black header while it's focused produces something like this: 
Screenshot 2024-01-18 at 12 00 49

2. Similarly, hovering on the other one while it's focused looks like this (the colour is off and there is an additional line above the black one):
Screenshot 2024-01-18 at 12 05 39
Comparing with the "Your services" and "Security" links which look like this when focused+hovered:
Screenshot 2024-01-18 at 12 05 49
We should probably try to keep all of them the same.  

3. There is still a problem with the spacing under the account navigation in some mobile browsers. Screenshot is from Chrome Browser on the latest iOS. 
Screenshot 2024-01-18 at 12 18 38

danacotoran:
I made a branch off your branch to experiment with this problem and look into the issues you have been trying to fix all along, to do with the additional spacing being added under the nav and the unexplained difference in underline thickness between the other links and the "Sign out" button. 
I _think_ I managed to find a rough solution that avoids using pseudo-classes and adding too much custom CSS trying to replicate the look and behaviour of a link.

Here is [a diff between your branch and the copy I created](https://github.com/govuk-one-login/di-account-management-frontend/compare/OLH-1224-sign-out-link-as-post-request...testy-button-stuff-testy) if you are interested.

danacotoran:
I thought I had submitted this comment. 
I had, in fact, closed the tab without submitting it, and lost all the content I had added. 🤦🏻‍♀️ So had to re-write it all.

Thanks for persevering with this. Unfortunately still found a couple of UI issues but seems like it's getting closer to being done!
- the sign out link in the account navigation does not have a hover style, and the focus style is different from the others. It also doesn't change when the `account-navigation__list-item--active` modifier is applied


https://github.com/govuk-one-login/di-account-management-frontend/assets/7116819/79697af6-e262-422e-b4b9-5a27d9d76290

- the link in the black header looks much better now, but a white underline still appears when it's hovered while also being in the focus state 

https://github.com/govuk-one-login/di-account-management-frontend/assets/7116819/f906a9e4-b0f3-40ae-96e2-15ee1c3e82fd

I see you went with the pseudo element route to apply the underline to the elements on hover. While this produces something looking very similar to the styling on a regular link, I think you should be able to use the same hover styles as regular design system links have (`text-decoration` based styles) so long as they are applied to the text level content of the button instead of the button itself.

Let me know if anything doesn't make sense or if there is anything I can do to help.

src/assets/scss/application.scss
```diff
@@ -197,11 +197,81 @@ $govuk-new-link-styles: true;
...
 $nav-link-outline-thickness: 2px;
 $nav-active-border-thickness: 5px;
 
-.account-navigation__link {
+/* Combined styles for buttons */
+.account-header-navigation__button,
+.account-navigation__button {
+  border: 0 none;
+  cursor: pointer;
+  background-color: unset;
```
danacotoran, line 209: My suggestion would be to change this to `background-color: transparent` which should produce the same outcome, but has better cross-browser support, so the "Sign out" link would be readable when viewed on a browser that doesn't support `unset`, whereas right now it's not:
unsetie11

danacotoran (commented):
Hey @aenciso thanks for this, the sign out link in the black header looks perfect (I only just have one tiny suggestion – I will pop that in as an inline comment).

> Not sure if you remember but in some of the earlier comments we said that there's no point of using because the user is redirected anyway and will never have this item active.

Yeah to be fair that's true, it seems like only "Your services" and "Security" have the ability to be "active". 
https://github.com/search?q=repo%3Agovuk-one-login%2Fdi-account-management-frontend%20activeNav&type=code The "Sign out" one might need to have an active state if we ever introduce an interstitial "Are you sure you want to sign out" type screen. So perhaps we can ignore that point for now. 

On the secondary nav, I spotted that the height of the nav is larger now, possibly as a consequence of increasing the padding on the links. Not sure whether this is something you discussed with Anisa that you are implementing deliberately, or whether it's by accident. 

barheight

I also noticed a discrepancy in focus states which you can hopefully see in the video: the focus state of the Sign out link does not align perfectly, it's like there is a 1px or so deviation. 

https://github.com/govuk-one-login/di-account-management-frontend/assets/7116819/f97535a2-a41f-453d-a0fd-c8e692c89f87

When it's both hovered and focused, the colour of the text changes slightly from black to a very dark blue (this is quite hard to spot). 


https://github.com/govuk-one-login/di-account-management-frontend/assets/7116819/482d1f33-5536-4989-ba73-797c454c76d5

Although, confusingly enough, this does not happen when I run your branch locally, but I can see it on the version of your branch that I deployed to dev (same browser). 🤨
I think what is happening is [this rule](https://github.com/govuk-one-login/di-account-management-frontend/pull/1139/files#diff-836f717cf44ec5735f83f3491d1e3313c2f9ae2963b7e04e5dcedc125b5bc4a6R253) taking precedence over the colour set by `govuk-focused-text` [here](https://github.com/govuk-one-login/di-account-management-frontend/pull/1139/files#diff-836f717cf44ec5735f83f3491d1e3313c2f9ae2963b7e04e5dcedc125b5bc4a6R253).

Let me know if you have questions or if you could use some help with any of that.

danacotoran (commented):
Thanks for all the progress on this @aenciso. Hopefully not much longer. 
Most of the cross-browser related issues seem to have been resolved which is good.
After another fairly thorough round of testing, I could spot one small issue we should fix, where the account nav "jumps" a little bit – hopefully you can see what I mean from the recording (it's from the latest version of Chrome browser on my laptop, if that helps. But it's not only happening on Chrome). 

https://github.com/govuk-one-login/di-account-management-frontend/assets/7116819/03f82ffd-c01a-493a-8af5-33cdacc2c7d3

danacotoran:
@aenciso I've given this another look: the nav is definitely still shifting when "Sign out" is in focus, however since this 1. looks good otherwise 2. is blocking another major piece of work and 3. it's one of those tricky cross-browser things, let's call this good enough for the time being.

Would you be able to create a new backlog ticket (linking it to the existing one) to investigate and fix the "jumping" nav problem in Chrome and Firefox?
And would you mind cleaning up the commit history on this branch as well?

Happy to approve once that has been done.

danacotoran (approved):
Thanks for the perseverance with this one 👍

PR #1145, opened 2024-01-08
OLH-1336 Use a singleton for the SQS client.

andrew-moores:
## Proposed changes

Change the frontend so that it doesn't create a new SQS client every time it needs to publish an event.

### What changed

Introduced a new singleton SQS client object.

### Why did it change

In prod we are seeing CredentialsProviderError which we think is due to the existing pattern of creating a new SQS client every time an event is published.  

## Testing

It is deployed in dev.  I have visited the triage page several times and seen audit events published each time.  I visited the page over a period greater than an hour to check that the temporary creds that the client obtains are refreshed.

The error was seen in the build environment when running a performance test against the triage page.  To gain confidence this fixes the issue we could run the perf test over a longer period than previously run.

### DIFF
--- a/src/config/aws.ts
+++ b/src/config/aws.ts
@@ -6,7 +6,7 @@ import {
   getKmsKeyId,
   getLocalStackBaseUrl,
 } from "../config";
-import { SQSClientConfig } from "@aws-sdk/client-sqs";
+import { SQSClient, SQSClientConfig } from "@aws-sdk/client-sqs";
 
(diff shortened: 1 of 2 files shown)


alex9smith:
I've been trying to think if there's any tests we could add with this change, but we already have tests at the contact controller that check we're sending an audit event to SQS. I'm not sure there's much value in adding an integration test for this service

PR #1149, opened 2024-01-15
[DO NOT MERGE] OLH-1263: Bump govuk-frontend to v5

danacotoran:
### What changed
Implement changes required to update `govuk-frontend` to the latest version which is a breaking one, removing JS support for <IE11, introducing breaking updates to the JS API, amongst other changes.
- update SASS/JS/asset load paths to include `/dist`: everything in the package has been moved to a `dist` folder
- update the way components are initialised and imported – the `GOVUKFrontend` window global has been removed, plus you no longer need to call the `init` method to initialise a component
- bundle component dependencies from GOVUK Frontend into `application.js` – I did this using `rollup` but if anyone has opinions feel free to share
- reformat showPassword component to match the new format of Design System components
- remove `$govuk-new-link-styles` flag from the application sass file as the new styles are no longer opt-in only
- load cookies script separately as a normal script while all other application JS is loaded as a `<script type="module">`
-  remove ie8 conditional from header
- remove `<span class="govuk-header__logotype-text">` and its content from the header and update the SVG to the supplied version
- remove `span` wrapper around the `svg` in the header

### Why did it change
The latest update to the package introduces breaking changes

## Testing
Tested manually in dev and on local, checking that components are working, cookie functionality is preserved, and `application.js` contains only scripts of components we are using in account management. 

## How to review 

Please check the [release notes for v5](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0) and highlight if anything isn't covered in this PR.
Check that no breaking changes were introduced as a result of this update, for instance:
- components are still styled correctly
- links are styled correctly
- no essential functionality is missing
- cookies/analytics code still running as expected 
- JS code bundles correctly and is uglified for the production build


benjamin-mortimer:
Please hold off on merging this - still checking with the UCD leads about the visual changes.

PR #1150, opened 2024-01-16
Olh 1337 redact user id from txma error log

andrew-moores:
## Proposed changes

Improve the way errors sending events to TxMA are handled.

### What changed

Introduced a DLQ to put TxMA events on if they fail sending to the TxMA queue.
Introduced a redact function to remove sensitive data from json before logging.
Used the redact function to ensure we don't log the user id in the TxMA event in the scenario where we can't publish the event to any queues.

### Why did it change

We need to make sure we don't lose TxMA events when we can't send them to the TxMA queue.  Sending them to a DLQ allows us to retry the send.

We cannot log the user id in messages that are visible on the performance index in production.

## Testing

During development of this change I noticed that the tests in the test/unit folders were being excluded from the npm run test:unit script.  Some of these tests were for modules changed for the PR and they were moved into the src folder structure to ensure they get run.


peterfajemisincabinetoffice (approved):
LGTM

alex9smith (approved):
This looks good to me too 👍🏻

PR #1153, opened 2024-01-17
[OLH-1317] Reduce number of logging messages by excluding the healthcheck

saralk:
## Proposed changes

This adds the /healthcheck route to the list of paths for the logger to ignore.

### Why did it change

The log message does not add any real value and clutters up the logs.

## Testing

I have run this locally; and verified that without the change, a log item appears when the endpoint is hit. And with the change, the log item does not appear.

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -54,6 +54,7 @@ const loggerMiddleware = PinoHttp({
       "/assets/fonts/bold-b542beb274-v2.woff2",
       "/assets/images/favicon.ico",
       "/assets/fonts/light-94a07e06a1-v2.woff2",
+      "/healthcheck",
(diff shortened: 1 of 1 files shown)


src/utils/logger.ts
```diff
@@ -54,6 +54,7 @@ const loggerMiddleware = PinoHttp({
       "/assets/fonts/bold-b542beb274-v2.woff2",
       "/assets/images/favicon.ico",
       "/assets/fonts/light-94a07e06a1-v2.woff2",
+      "/healthcheck"
```
andrew-moores, line 57:
```suggestion
      "/healthcheck",
```

andrew-moores (approved):

PR #1155, opened 2024-01-18
OLH-1363: Custom nonce approach for webchat styles

danacotoran:
### What changed
The webchat supplier have updated the webchat so that it will work with our content security policy.

We can now include the webchat stylesheets manually into our code with the nonce generated by us. This implements that change, adding the stylesheets to the `<head>` conditionally only when webchat is enabled, and storing the paths as an environment variable.

The `href` values will be different per environment but the webchat supplier have not supplied the values for other environments yet, so it's only dev for now.
Once they have confirmed the values for other environments, the environment variables can be updated.

Note: the supplier’s script is still injecting the styles too so we end up with two sets of the same (or similar?) stylesheets (the ones with `contact-gov-uk-one-login` as initiator in the Network tab are the new inline ones we added, while the others are originating from the supplier’s script).
Also, adding the new inline stylesheets is causing the webchat button icon to display a bit mis-centred. I haven't quite figured out why but it might be one to investigate separately.

Screenshot 2024-01-17 at 10 22 16

### Why did it change
This approach allows us to remove the 'unsafe-inline' keyword from the `style-src` directive and is more in line with our content security policy.

### Related links
https://govukverify.atlassian.net/browse/OLH-1363

## Testing
Tested manually on local with the feature flag on/off. Checked for console errors and network errors.

## How to review
Check that everything is correct with the new environment variables, and that the new code is added only when webchat is on. Confirm that there are no CSP errors (or other types of error) resulting from these changes.


alex9smith (approved):
Nice one, thanks Dana

PR #1156, opened 2024-01-18
OLH-1368 - Onboard Veterans Card.

andrew-moores:
## Proposed changes

[OLD-1368] Add support for HM Armed Forces Veteran Card

### What changed

- Add `veteransCard` as an allowed clientID
- Show a link to the veterans card page when the user has that in their recent activity
- Display veterans card link in the "Services you can use with GOV.UK One Login" List
- Add Welsh translation

### Why did it change

The Veterans Card service is going live soon, and we need to support it

## Testing

Verified that the link shows when the activity is present, and that the links work

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -151,6 +151,8 @@ export const getAllowedServiceListClientIDs: string[] = [
   "dbs",
   "vehicleOperatorLicense",
   "mortgageDeed",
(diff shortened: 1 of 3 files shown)


alex9smith (approved):
LGTM

PR #1161, opened 2024-01-25
[OLH-1299] Query for encrypt and decrypt all user events

aenciso:
## Proposed changes

Update the way we access activity_log data directly from the frontend

### What changed

The database model has changed so instead of having an array of activities it now return one activity per row

### Why did it change

The activity log table is no longer accessed through lambda and can now be accessed directly from the frontend 

### Related links

https://govukverify.atlassian.net/browse/OLH-1299

https://github.com/govuk-one-login/di-account-management-backend/blob/main/docs/adr/0010-simplify-activity-log-data-structure-pipeline.md

- [ ] GENERATOR_KEY
- [ ] WRAPPING_KEY


deploy/template.yaml
```diff
@@ -102,11 +102,6 @@ Parameters:
   ActivityLogStoreTableName:
     Type: String
     Default: activity_logs
-  CodeSigningConfigArn:
```
aenciso, line 105: We are not using this anywhere

deploy/template.yaml
```diff
@@ -123,8 +118,6 @@ Conditions:
   UseStubServices: !Or
     - !Equals [!Ref Environment, build]
     - !Equals [!Ref Environment, dev]
-  UseCodeSigning:
```
aenciso, line 126: We are not using this anywhere

PR #1172, opened 2024-02-18
OLH-1299 Decrypt user events

aenciso:


src/components/change-email/tests/change-email-integration.test.ts
```diff
@@ -17,6 +17,7 @@ describe("Integration:: change email", () => {
   let token: string | string[];
   let cookies: string;
   let app: any;
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
andrew-moores, line 20: Is this needed?  When I remove it locally I don't get any linting errors?

src/utils/decrypt-data.ts
```diff
@@ -0,0 +1,95 @@
...
+const decryptClient = buildDecrypt(decryptClientConfig);
+
+let keyring: KmsKeyringNode;
+
+export async function generateExpectedContext(
+  userId: string
+): Promise<EncryptionContext> {
+  const { AWS_REGION, ACCOUNT_ID, ENVIRONMENT, VERIFY_ACCESS_VALUE } =
+    process.env;
+  if (AWS_REGION === undefined) {
```
andrew-moores, line 24: Consider using node assert in a try catch to improve readability of these guard conditions.

src/utils/decrypt-data.ts
```diff
@@ -0,0 +1,95 @@
...
+    logger.error("Missing ENVIRONMENT environment variable");
+  }
+
+  if (VERIFY_ACCESS_VALUE === undefined) {
+    logger.error("Missing VERIFY_ACCESS_VALUE environment variable");
+  }
+
+  let accessCheckValue;
+  try {
+    accessCheckValue = await getHashedAccessCheckValue(VERIFY_ACCESS_VALUE);
```
andrew-moores, line 42: What is VERIFY_ACCESS_VALUE?

src/utils/decrypt-data.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export function validateEncryptionContext(
+  context: EncryptionContext,
+  expected: EncryptionContext
+): void {
+  if (context === undefined || Object.keys(context).length === 0) {
+    logger.error("Encryption context is empty or undefined");
+  }
+
+  Object.keys(expected).forEach((key) => {
```
andrew-moores, line 64: Is it a problem if the context has more keys than the expected context?

src/utils/decrypt-data.ts
```diff
@@ -0,0 +1,95 @@
...
+      keyring,
+      Buffer.from(data, ENCODING)
+    );
+    validateEncryptionContext(
+      result.messageHeader.encryptionContext,
+      await generateExpectedContext(userId)
+    );
+    return result.plaintext.toString(DECODING);
+  } catch (error) {
+    logger.error(error, "Failed to decrypt data.");
```
andrew-moores, line 92: Does that log everything in the error?

andrew-moores, line 92: Should we include the trace id in this log message so we can determine which user it impacted?

src/utils/test/decrypt-data.test.ts
```diff
@@ -0,0 +1,213 @@
...
+  });
+
+  it("should log an error if decryption fails", async () => {
+    decryptStub.rejects(new Error("Decryption failed"));
+    await decryptData("encryptedString", "testUser");
+    expect(loggerErrorSpy.calledWith("Failed to decrypt data.")).to.be.true;
+  });
+
+  // Add more tests as needed to cover different scenarios
+});*/
```
andrew-moores, line 213: If we don't need these tests they can be deleted.

src/utils/types.ts
```diff
@@ -25,20 +25,24 @@ export interface KmsService {
...
+  event_type: string;
+  reported_suspicious: boolean;
+  session_id: string;
 }
 
-export interface Activity {
-  type: string;
-  client_id: ClientId;
-  timestamp: number;
+export interface FormattedData {
```
andrew-moores, line 37: This name isn't very informative.  Could it be changed to FormattedActivityLog?  Was this copied from the report suspicious activity branch?

Dockerfile
```diff
@@ -5,7 +5,10 @@ COPY package-lock.json ./
 COPY tsconfig.json ./
 COPY ./src ./src
 COPY ./@types ./@types
-RUN npm install && npm run build && npm run clean-modules && npm install --production=true
+RUN npm install
+RUN npm run build
+RUN npm run clean-modules
+RUN npm install --production=true
```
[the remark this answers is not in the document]
aenciso: Yes, you are right. I split them because it was failing in dev and I wasn't getting any useful information. I'll revert this change

src/utils/types.ts
aenciso: Done

src/utils/decrypt-data.ts
aenciso: Done

src/utils/decrypt-data.ts
aenciso: The stacktrace yes

aenciso: I think it's useful yes. I'll add that

aenciso: Done

src/utils/decrypt-data.ts
andrew-moores: Yes it is a problem so logging an error is correct. 👍

andrew-moores (approved):

PR #1176, opened 2024-02-21
OLH-1505: Refactor date formatting function

danacotoran:
## Proposed changes
The `prettifyDate` util was hard coded to return dates in `en-GB`. There is a need to display dates in Welsh when the user is viewing their OL home in Welsh.
This refactors the formatter to allow an additional locale parameter to be passed in, to display dates in Welsh when the language has been set to Welsh.

### What changed
Extend `prettifyDate` to accept a language parameter which should be one of 'en' or 'cy'. 
If language is not specified, or the parameter has a value other than 'en' or 'cy', default to returning dates in English. 
Refactor the parts of the code which rely on `prettifyDate`, ensuring a language parameter is passed.
`prettifyDate` now accepts named parameters.

### Why did it change
Welsh language support is a requirement.

### Related links
https://govukverify.atlassian.net/browse/OLH-1505

## Testing
Automated testing and manual testing of `prettifyDate` passing different options

## How to review
Code review and reviewing the different pages that rely on the `prettifyDate` function, in English and Welsh

### DIFF
--- a/src/app.constants.ts
+++ b/src/app.constants.ts
@@ -132,6 +132,15 @@ export enum LOCALE {
   CY = "cy",
 }
 
+export interface LanguageCodes {
+  en: 'en-GB',
+  cy: 'cy-CY'
+}
+export const LANGUAGE_CODES: LanguageCodes = {
+  en: "en-GB",
+  cy: "cy-CY",
+}
+
 export const LOG_MESSAGES = {
   ATTEMPTING_TO_DESTROY_SESSION: "Attempting to destroy session",
   GS_COOKIE_NOT_IN_REQUEST: "gs cookie not in request.",
--- a/src/utils/prettifyDate.ts
+++ b/src/utils/prettifyDate.ts
@@ -1,5 +1,15 @@
-export const prettifyDate = (dateEpoch: number, options?: Record<string, unknown>): string => {
-  const params = options || { dateStyle: "long" }
+import { LANGUAGE_CODES } from "../app.constants";
+
+export const prettifyDate = ({ 
+    dateEpoch, 
+    locale, 
+    options = { dateStyle: "long" }
+  }: {
+    dateEpoch: number,
+    locale?: string,
+    options?: Intl.DateTimeFormatOptions
+  }): string => {
+  const languageCode = LANGUAGE_CODES[locale as keyof typeof LANGUAGE_CODES] || "en-GB";
   let dateEpochInMilliSeconds: number;
   if (dateEpoch.toString().length === 10) {
     // Epoch is in seconds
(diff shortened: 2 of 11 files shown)


src/utils/signInHistory.ts
```diff
@@ -118,14 +118,18 @@ export const formatData = (
...
+      options: {
+        month: "long",
+        day: "numeric",
+        year: "numeric",
+        hour: "numeric",
+        minute: "numeric",
+        hourCycle: "h12",
+        timeZone: "GB",
+      },
+      locale: currentLanguage
```
saralk, line 132: Should we default to `en-gb` if the language is valid, but a non supported language (e.g. `fr-fr`)

danacotoran: The way I handled that scenario (or attempted to) in `prettifyDate` is by checking whether the code that's been passed through is a key in the list of `LANGUAGE_CODES` that I added (currently we only support `cy` and `en` translations so I mapped those two to their corresponding language tags) and falling back to `en-GB` if it doesn't match.

Not sure whether my approach to the problem was the the most optimal one but if you have any feedback, I'm open to suggestions!

src/app.constants.ts
```diff
@@ -132,6 +132,11 @@ export enum LOCALE {
   CY = "cy",
 }
 
+export const LANGUAGE_CODES: Record<string, string> = {
```
aenciso, line 135: I think we could make use of an enum or an interface here to make it less generic perhaps?
`export interface LanguageCodes {
  en: 'en-GB',
  cy: 'cy-CY'
}

export const LANGUAGE_CODES: LanguageCodes = {
  en: "en-GB",
  cy: "cy-CY",
}`

src/utils/prettifyDate.ts
```diff
@@ -1,5 +1,15 @@
...
+import { LANGUAGE_CODES } from "../app.constants";
+
+export const prettifyDate = ({ 
+    dateEpoch, 
+    locale, 
+    options = { dateStyle: "long" }
+  }: {
+    dateEpoch: number,
+    locale?: string,
+    options?: Record<string, unknown>
```
aenciso, line 10: We can use Intl.DateTimeFormatOptions instead to make it more strict

src/utils/signInHistory.ts
```diff
@@ -99,11 +99,11 @@ export const generatePagination = (dataLength: number, page: any): [] => {
 
 export const formatData = (
   data: ActivityLogEntry[],
-  currentPage?: number
+  currentLanguage: string,
```
aenciso, line 102: I think this should be treated as optional even though we need it as req.i18n.language is optional. This makes sure that we do all checks before we use it

src/utils/prettifyDate.ts
danacotoran: Nice suggestion, have done!

aenciso (approved):

alex9smith (approved):

PR #1177, opened 2024-02-23
OLH-1382 - report suspicious activity.

andrew-moores:
## Proposed changes

Implement the link between the activity history page and the report/reported activity history page.

### What changed
The link from the activity history page to the reporting page was implemented using path parameters.  The report page retrieves the activity from DynamoDB and performs various checks on the query parameters.  If the activity has already been reported then the already reported paged is displayed.

Changes also made the provision script and the SAM template.

### Why did it change

To allow users to report suspicious activity and to help developers test this feature.

## Testing

Can be deployed to dev.


src/components/activity-history/tests/activity-history-integration.test.ts
```diff
@@ -0,0 +1,306 @@
...
+  });
+
+  it("should display without pagination when data is less than activityLogItemsPerPage", async () => {
+    const dataShort = [
+      {
+        event_type: "signed-in",
+        session_id: "asdf",
+        user_id: "string",
+        timestamp: "1689210000",
+        activities: [
```
aenciso, line 57: I think these tests should reflect the new activity_log table structure

src/components/report-suspicious-activity/report-suspicious-activity-controller.ts
```diff
@@ -0,0 +1,124 @@
...
+      try {
+        assert.strictEqual(req.query.reported.toString(), "false");
+        alreadyReported = false;
+      } catch (err) {
+        req.log.error(
+          `Query parameter "reported" was <${req.query.reported}> but expected "true" or "false" `
+        );
+        return next();
+      }
+    }
```
saralk, line 51: Why do we need the the `reported` query parameter? Can we not get the value from the DB and render accordingly?

If we do need the query parameter, then this code is a bit hard to read, could something like this work...
      
```suggestion
    assert.match(req.query.reported.toString(), /^(true|false)$/, `Query parameter "reported" was <${req.query.reported}> but expected "true" or "false" `);
    alreadyReported = req.query.reported.toString() === "true"
```

src/components/report-suspicious-activity/report-suspicious-activity-controller.ts
```diff
@@ -0,0 +1,124 @@
...
+        `Event ${req.query.event} matched ${response.Items.length} Items when 1 was expected.`
+      );
+      return next();
+    }
+    eventDetails = DynamoDB.Converter.unmarshall(
+      response.Items[0]
+    ) as ActivityLogEntry;
+    assert.ok(
+      eventDetails.reported_suspicious === alreadyReported,
+      `event showed reported_suspicious as ${eventDetails.reported_suspicious} but request param reported was ${alreadyReported}`
```
saralk, line 73: Same as above, do we need this? Can we not just render based on what's in the database?

src/components/activity-history/tests/activity-history-integration.test.ts
aenciso: If it's too much hassle, perhaps we can wait after we merge https://github.com/govuk-one-login/di-account-management-frontend/pull/1172 or perhaps borrow this approach? whichever is easier https://github.com/govuk-one-login/di-account-management-frontend/pull/1172/files#diff-ef3df6a5514d9bc94baf4d3f411b0c9d6efc7d0aa16d1a6d54496afa90399237R85-R101

andrew-moores: I've changed the integration tests so they reflect the correct structure of the log and fixed a bug where empty rows could have been displayed on the activity history.

aenciso (approved):

peterfajemisincabinetoffice (approved):

saralk (approved):

src/components/report-suspicious-activity/report-suspicious-activity-routes.ts
```diff
@@ -0,0 +1,23 @@
...
+import { reportSuspiciousActivityGet, reportSuspiciousActivityPost } from "./report-suspicious-activity-controller";
+import { PATH_DATA } from "../../app.constants";
+import { asyncHandler } from "../../utils/async";
+import { requiresAuthMiddleware } from "../../middleware/requires-auth-middleware";
+import { refreshTokenMiddleware } from "../../middleware/refresh-token-middleware";
+
+const router = express.Router();
+
+router.get(
+  PATH_DATA.REPORT_SUSPICIOUS_ACTIVITY.url,
```
saralk, line 11: was just thinking about this, instead of using a query parameter, could we do something like
```suggestion
  `${PATH_DATA.REPORT_SUSPICIOUS_ACTIVITY.url}/:event_id`,
```
It wouldn't make a huge difference, but the URLs would be a bit nicer looking, and maybe a bit more semantic?

src/components/report-suspicious-activity/report-suspicious-activity-controller.ts
```diff
@@ -0,0 +1,106 @@
...
+): Promise<void> {
+  try {
+    assert.ok(
+      req.query.event !== undefined && req.query.event !== "",
+      "mandatory query parameter 'event' missing from request"
+    );
+  } catch (err) {
+    req.log.error(err.message);
+    return next();
+  }
```
saralk, line 36: If we used a route parameter, we wouldn't need this

src/components/report-suspicious-activity/report-suspicious-activity-routes.ts
andrew-moores: There are more changes than this required to do this so will postpone until next ticket where I will look at the urls again.

src/components/report-suspicious-activity/report-suspicious-activity-controller.ts
andrew-moores: Same as for last comment.

saralk (approved):

PR #1179, opened 2024-02-27
OLH-1456: Add HMRC card to the service dashboard

danacotoran:
### What changed
Add content for the HMRC service card on the service dashboard. The card will live under "Other services you've used" as opposed to "Your accounts".

Screenshot 2024-02-27 at 13 35 20
Screenshot 2024-02-27 at 13 35 53

It has a different design from the other services on the service dashboard due to the requirement to have more content to explain to users what the card is about. For the time being I handled this by adding an "isHMRC" flag which is true if the id of a service matches one of the HMRC client IDs.

For now we should not add this service to the "Services you can use" list as they'll be in private beta initially.

### Why did it change
HMRC is onboarding with OL at the end of Feb (initially as private beta) so they need a service card. 

## Testing
Tested locally and in dev, in isolation and with other service cards, in English and Welsh

### DIFF
--- a/src/components/your-services/index.njk
+++ b/src/components/your-services/index.njk
@@ -55,10 +55,23 @@
           {% set locale = ['clientRegistry.', env, '.', service.client_id, '.'] | join %}
           <li class="your-services__card">
             <div class="your-services__card__content">
-            <{{serviceHeadingLevel}} class="govuk-heading-s govuk-!-font-weight-regular" data-test-id="service-card-heading">
-              <a class="govuk-link" href="{{ [locale, 'link_href'] | join | translate }}">{{ [locale, 'link_text'] | join | translate }}</a>
-            </{{serviceHeadingLevel}}>
-            <p class="govuk-body your-services__card__last-used">{{ 'pages.yourServices.lastUsed' | translate }}: {{ service.last_accessed_readable_format }}</p>
+            {% if service.isHMRC %}
+            {# The HMRC service card contains more information than the regular ones #}
+              <{{serviceHeadingLevel}} class="govuk-heading-s" data-test-id="service-card-heading">
+                {{ [locale, 'header'] | join | translate }}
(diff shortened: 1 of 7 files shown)


src/utils/yourServices.ts
```diff
@@ -81,11 +82,13 @@ export const getAllowedListServices = async (
...
   const readable_format_date = prettifyDate(service.last_accessed);
+  const isHMRC = hmrcClientIds.includes(service.client_id);
   return {
     client_id: service.client_id,
     count_successful_logins: service.count_successful_logins,
     last_accessed: service.last_accessed,
     last_accessed_readable_format: readable_format_date,
+    isHMRC: isHMRC,
   };
 };
```
andrew-moores, line 94: Should we have included this in a unit test?  Does this change reduce the Sonarqube coverage score?

src/components/your-services/index.njk
```diff
@@ -55,10 +55,23 @@
           {% set locale = ['clientRegistry.', env, '.', service.client_id, '.'] | join %}
           <li class="your-services__card">
             <div class="your-services__card__content">
-            <{{serviceHeadingLevel}} class="govuk-heading-s govuk-!-font-weight-regular" data-test-id="service-card-heading">
-              <a class="govuk-link" href="{{ [locale, 'link_href'] | join | translate }}">{{ [locale, 'link_text'] | join | translate }}</a>
-            </{{serviceHeadingLevel}}>
-            <p class="govuk-body your-services__card__last-used">{{ 'pages.yourServices.lastUsed' | translate }}: {{ service.last_accessed_readable_format }}</p>
+            {% if service.isHMRC %}
```
saralk, line 58: do we need this? can we not just check if there is a `paragraph1` field? Would make it more semantic

andrew-moores (approved):

PR #1187, opened 2024-03-08
OLH-1521 Make activity log and RSA available only if the user has certain service cards

aenciso:
[OLH-1521] Make activity log and RSA available only if the user has certain service cards

## Proposed changes

Only users with HMRC service in their list would be able to access the activity history page and report suspicious activity page

### What changed

Added checks to verify if the user has HMRC services in their list

### Why did it change

To reduce the risk of overwhelming TICF.

## Testing

Unit, integration and local testing

[OLH-1521]: https://govukverify.atlassian.net/browse/OLH-1521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/activity-history/activity-history-routes.ts
+++ b/src/components/activity-history/activity-history-routes.ts
@@ -2,12 +2,14 @@ import * as express from "express";
 import { activityHistoryGet } from "./activity-history-controller";
 import { PATH_DATA } from "../../app.constants";
 import { requiresAuthMiddleware } from "../../middleware/requires-auth-middleware";
+import { checkAllowedServicesList } from "../../middleware/check-allowed-services-list";
 
 const router = express.Router();
 
 router.get(
   PATH_DATA.SIGN_IN_HISTORY.url,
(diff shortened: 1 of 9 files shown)


alex9smith (approved):
Nice, this is really clean as middleware. Will we need to add some data for our user in dev so they can still see the activity history?

aenciso:
> Nice, this is really clean as middleware. Will we need to add some data for our user in dev so they can still see the activity history?

Ah yes, we need to add HMRC to the activity logs for our dev user. Let me do that. As part of this ticket.

saralk:
I've created a [new route](https://github.com/govuk-one-login/di-account-management-frontend/blob/main/src/components/report-suspicious-activity/report-suspicious-activity-routes.ts#L27-L31), could you rebase and add this middleware to that route?

PR #1189, opened 2024-03-11
OLH-1565 Redirect the user after submitting a suspicious activity report.

saralk:
## Proposed changes

Add a redirect to the POST handler for suspicious activity, to redirect the user to the confirmation page.

### What changed

- Add a new GET handler for the `report-activity/done` route
- In the POST handler, redirect the user to the GET handler.
- Ensure the `page` parameter is added to the handler correctly.

### Why did it change

During testing, we discovered an issue where if a user reports activity, then clicks on the browser's back button; they would see a "Confirmation of resubmitting form" message. This stops that from happening. By using a 302 redirect, the POST page is not included in the history, so it wont be resubmitted when the user goes back.

## Testing

I have tested this locally, and on dev.

### DIFF
--- a/src/components/report-suspicious-activity/report-suspicious-activity-controller.ts
+++ b/src/components/report-suspicious-activity/report-suspicious-activity-controller.ts
@@ -154,8 +154,12 @@ export async function reportSuspiciousActivityPost(
     return next(err);
   }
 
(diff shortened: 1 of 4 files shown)


alex9smith (approved):
👍🏻 We'll want to add the middleware added in #1187 to this new route but let's do that as a separate PR once all the changes are in

PR #1192, opened 2024-03-12
OLH-1561 Add initial tranche of RPs to RSA allow list.

andrew-moores:
## Proposed changes

Add additional RPs to the activity history allow list.

### What changed

Updated the constants file to include a new allow list for RSA.
Introduced some constants for the RP client ids.

### Why did it change

For go live of the RSA feature we want a selection of RPs users to be able to acces it.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -119,6 +119,19 @@ export function getServiceDomain(): string {
   return process.env.SERVICE_DOMAIN ?? "";
 }
 
+const DBS_PROD: string = "Dw7Cxas8W7O2usHMHok95elKDRU";
+const DBS_NON_PROD: string = "dbs";
+const AAS_PROD: string = "dVrdJ7aemrvR0YlX7lDRaXnz0mE";
+const AAS_NON_PROD: string = "aas";
+const APAR_PROD: string = "2nAxHa72OqhE6eKymHZIx-sV3vI";
+const APAR_NON_PROD: string = "apar";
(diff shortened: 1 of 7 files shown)


src/config.ts
```diff
@@ -119,6 +119,19 @@ export function getServiceDomain(): string {
   return process.env.SERVICE_DOMAIN ?? "";
 }
 
+const DBS_PROD: string = "Dw7Cxas8W7O2usHMHok95elKDRU";
```
alex9smith, line 122: This is a great idea, makes our clunky solution much more manageable

alex9smith (approved):

peterfajemisincabinetoffice (approved):

src/middleware/check-allowed-services-list.ts
```diff
@@ -23,12 +23,23 @@ export const hasHmrcService = async (
...
+export async function checkRSAAllowedServicesList(
   req: Request,
   res: Response,
   next: NextFunction
 ): Promise<void> {
-  if (await hasHmrcService(req, res)) {
+  if (await hasAllowedRSAServices(req, res)) {
     next();
   } else {
     res.redirect(PATH_DATA.SECURITY.url);
```
aenciso, line 45: I forgot to add some logging here, maybe it's useful from a security perspective?

aenciso (approved):

PR #1193, opened 2024-03-12
OLH-1559: Add activity log start date

alex9smith:
### What changed

Add today's date to the translation files for English and Welsh where we display the "GOV.UK One Login started recording..." message.

### Why did it change

We released the backend to production this morning and are now storing data. This will explain to users why they don't see history before today. 

### Testing
I've tested this in dev in both English and Welsh
image
image


danacotoran:
To be fair we could also just hard code the date in the translation files since we now know it's `12 March 2024`/`12 Mawrth 2024`.  Saves us calling the formatting function again and doing the `replace` thing with Nunjucks, though there might be some advantage to this approach that I haven't thought of!

alex9smith:
Hmm yes I think you're right. I'm getting too clever. We're never going to want to change this date once it goes live. I'll get rid of all the code changes and put it in the translation files instead.

danacotoran (approved):
Nice one 🙌

PR #1198, opened 2024-03-13
OLH-1580: Sort activity history to display newest items first

alex9smith:
### What changed

Sort the activity history with the newest entry first.

We can't do this in the DynamoDB query because it only supports sorting using the sort key, and our sort key has to be the event ID because timestamp may not be unique. This means the entries come back from DynamoDB in an effectively random order.

The presenter is called before we do pagination, so sorting here means the order stays consistent across all the displayed pages.

This PR also moves the presenter into a separate file. This is so we can use `sinon` to stub the call to `getActivityLogEntry` and control the 

### Why did it change

Currently the activity history displays items in a random order.

## Testing

I've run this locally and checked the order is correct and consistent across pages (see below)

## How to review

I've put the refactor as its own commit to make it easier to review the change in logic.

Screenshot 2024-03-13 at 09 46 22

### DIFF
--- a/src/utils/activityHistory.ts
+++ b/src/utils/activityHistory.ts
@@ -118,7 +118,9 @@ export const formatActivityLog = (
   formattedActivityLog.clientId = activityLogEntry.client_id;
   formattedActivityLog.reportedSuspicious =
     activityLogEntry.reported_suspicious;
-  formattedActivityLog.reportSuspiciousActivityUrl = `${PATH_DATA.REPORT_SUSPICIOUS_ACTIVITY.url}?event=${activityLogEntry.event_id}&page=${pageNumber || 1}`;
+  formattedActivityLog.reportSuspiciousActivityUrl = `${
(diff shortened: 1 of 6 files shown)


src/utils/activityHistory.ts
```diff
@@ -188,7 +190,7 @@ const activityLogDynamoDBRequest = (
   ScanIndexForward: false, // Set to 'true' for ascending order
 });
 
-const getActivityLogEntry = async (
+export const getActivityLogEntry = async (
```
saralk, line 193: very small thing, but shouldn't this be called `getActivityLogEntries`

saralk (approved):

peterfajemisincabinetoffice (approved):

src/utils/activityHistory.ts
alex9smith: Yes, I agree. I'll update that in another PR later today.

PR #1199, opened 2024-03-13
Olh 1580 order activity log by timestamp using a lsi

andrew-moores:


alex9smith:
This was one of two options we were considering for this work and we went with the other. I'll close this PR but not delete the branch so we can come back to it if we ever want to switch

PR #1202, opened 2024-03-13
OLH-1583 Add the Stub RPs to the activity log allowlist

saralk:
### What changed

Auth run stub RP clients in all environments to allow testing of the One Login journey in all permutations. This PR adds the client IDs for these to our allow list of services that can see the activity log.

### Why did it change

This will give us an easy way of adding an allowed service to the history of our users in non-production environments for testing.


alex9smith (approved):

alex9smith:
Sonar are having an incident, so I'm using my admin powers to merge this PR even though the action is failing

PR #1209, opened 2024-03-15
OLH-1599 Update report suspicious activity content to remove contact route

andrew-moores:
## Proposed changes

Remove contact us route from RSA
### What changed

Page and email definitions.
### Why did it change
We can't support contact follow for RSA.

## Testing
Can be deployed to dev for a UCD review.
The updated templates require logging into Notify with the correct privs.  I also have copies that I sent to myself.


aenciso (approved):

danacotoran:
@andrew-moores is there any reason we've removed the references from the UI but kept the translations in the translation file? Do we foresee that we might need to add this content back in a future iteration?

PR #1216, opened 2024-03-20
OLH-1525: add tests for HMRC card

danacotoran:
### What changed
Address the comments on [this PR](https://github.com/govuk-one-login/di-account-management-frontend/pull/1179).

The points that were left unaddressed: 
1.  enquiring whether the conditional rendering of the longer service card design with `{% if service.isHMRC %}` can be made more semantic by checking whether a `paragraph1` field exists instead.
2. adding tests for the longer service card design

In this PR I addressed question 2. 

I attempted to address question 1, however I could not find a good way to conditionally display pieces of UI if a translation exists, or to determine whether a translation exists from the template layer in general. 

This doesn't seem to be something that we are currently doing anywhere else, and it doesn't seem straightforward to do, so I'm not sure whether it would warrant the effort that would be necessary to implement it (but let me know what you think).

It doesn't sound as if there are plans to use this longer style of card for any other service, however I did change the flag from `isHMRC` to `hasDetailedCard` to make it more generic. In the backend it's still HMRC specific (as in there is a check that determines whether the service client id is a HMRC one, and adds `hasDetailedCard` to the formatted data object if that is the case)

https://govukverify.atlassian.net/browse/OLH-1525 

### Why did it change
The changes had to go live before addressing the comments on the PR.
This addresses them retrospectively.

## How to review
A code review plus checking the HMRC card still displays correctly.

### DIFF
--- a/src/components/your-services/index.njk
+++ b/src/components/your-services/index.njk
@@ -53,9 +53,10 @@
       <ul class="govuk-list your-services__list">
         {% for service in servicesList %}
           {% set locale = ['clientRegistry.', env, '.', service.client_id, '.'] | join %}
-          <li class="your-services__card">
+          {% set testIdValue = "service-card-long" if service.hasDetailedCard else "service-card-short" %}
+          <li class="your-services__card" data-test-id={{testIdValue}}>
             <div class="your-services__card__content">
-            {% if service.isHMRC %}
+            {% if service.hasDetailedCard %}
             {# The HMRC service card contains more information than the regular ones #}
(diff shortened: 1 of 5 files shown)


andrew-moores:
I don't think checking for a specific piece of localisation configuration being defined is a good way to determine what should be displayed.  The name of the localisation property in this case is also pretty poor in conveying what it is we are testing for.  The rest of this change looks good to me.

andrew-moores (approved):

PR #1230, opened 2024-03-26
OLH-1636 - Error Recording Activity history

peterfajemisincabinetoffice:
### What changed
OLH-1636 - Error Recording Activity history
Remove refreshMiddlewareToken from RSA route.

### Why did it change

We discovered the refreshMiddlewareToken potentially always fails. 

### Related links

https://govukverify.atlassian.net/browse/OLH-1636

## Testing

Deploy to dev and ensure RSA journey is not broken

### DIFF
--- a/src/components/report-suspicious-activity/report-suspicious-activity-routes.ts
+++ b/src/components/report-suspicious-activity/report-suspicious-activity-routes.ts
@@ -7,7 +7,6 @@ import {
 import { PATH_DATA } from "../../app.constants";
 import { asyncHandler } from "../../utils/async";
 import { requiresAuthMiddleware } from "../../middleware/requires-auth-middleware";
-import { refreshTokenMiddleware } from "../../middleware/refresh-token-middleware";
 import { checkRSAAllowedServicesList } from "../../middleware/check-allowed-services-list";
 
 const router = express.Router();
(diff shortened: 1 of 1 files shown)


andrew-moores:
@peterfajemisincabinetoffice can you add an explanation of why we don't need this piece of middleware?  Its used throughout the code base so why is removing it from just this module the correct solution?

peterfajemisincabinetoffice:
> @peterfajemisincabinetoffice can you add an explanation of why we don't need this piece of middleware? Its used throughout the code base so why is removing it from just this module the correct solution?

Hi @andrew-moores , thanks for the call to discuss. 

Removing to unblock RSA as we think there is a wider issue with refreshing access token and will be dealt with as part of this tickets:

https://govukverify.atlassian.net/browse/OLH-1654
https://govukverify.atlassian.net/browse/OLH-1655
https://govukverify.atlassian.net/browse/OLH-1657

I have also carried out tests in dev environment and after a couple rounds of 10 mins, wait, 30 mins wait and over 1 hour wait, I can see there is no undesirable behaviour observed. Will need to merge to be able to carry out similar tests in higher environments as they might have different expiry times for access tokens.

andrew-moores (approved):

PR #1239, opened 2024-03-27
OLH-1316: focus order fix for accessibility

danacotoran:
### What changed

[OLH-1316: Fix webchat focus order issue](https://github.com/govuk-one-login/di-account-management-frontend/commit/7c374d75e1cb6d27496c5c673091b7f8865ea747) 

The purpose of these changes is to address an issue as described in the DAC audit report ([link in JIRA](https://govukverify.atlassian.net/browse/OLH-1316)). The issue is that focus order does not follow logically when the Webchat is activated using the "Use webchat" button in the content of the page (WCAG 2.4.3 Focus Order (Level A))

When navigating by keyboard or screen reader software, the following should happen:
- when the webchat is opened using the "Use webchat" button, focus should move to the webchat window (whereas currently users have to tab through the entire content of the page to reach the webchat window)
- when the webchat is closed using the "Use webchat" button, focus should move back to the "Use webchat" button (whereas currently focus remains on the floating button)

[Webchat asset conditional rendering](https://github.com/govuk-one-login/di-account-management-frontend/commit/587b9181fd7fee7daee210e9c053ff07227bfddc) 

Moved the `contactWebchatEnabled` condition into the `head` block on the webchat template. 
The reason for this change is I noticed that a request was being made for the stylesheet in production and other
environments where the webchat was off.
Moving this conditional block inside the head block instead of attempting to make the head block conditional upon `contactWebchatEnabled` being true seemed to fix the issue. It might have been a case of nunjucks being quirky.

Similarly: added the webchat related scripts to the scripts block which gets appended at the end of the body tag.

### Why did it change
These issues must be addressed before the webchat feature goes live. 

### Related links
Jira: https://govukverify.atlassian.net/browse/OLH-1316 
Accessibility report: https://drive.google.com/file/d/1drLiDgz5afi3rgyQot_eGl-ZN-wWEVLB/view?pli=1 

## Test
- the feature flag still works and no styles/scripts are loading when the flag is off
- the "Use webchat" button behaves as described above when navigating by keyboard

### DIFF
--- a/src/components/contact-govuk-one-login/index.njk
+++ b/src/components/contact-govuk-one-login/index.njk
@@ -6,13 +6,13 @@
 {% set languageToggle = true %}
 {% set secondaryContactMethodEnabled = contactPhoneEnabled or contactWebchatEnabled %}
 
-{% if contactWebchatEnabled %}
-  {% block head %} 
-    {{ super() }}
+{% block head %} 
+  {{ super() }}
+  {% if contactWebchatEnabled %}
     <link rel="stylesheet" nonce="{{scriptNonce}}" href="https://fonts.cdnfonts.com/css/rubik">
     <link rel="stylesheet" nonce="{{scriptNonce}}" href="{{webchatSource}}/css/loader.css">
-  {% endblock %}
-{% endif %}
+  {% endif %}
+{% endblock %}
 
 {% set emailBlock %}
   <section>
@@ -199,4 +199,5 @@
     }, 10000)
   </script>
 {% endif %}
+
 {% endblock %}
--- a/src/components/contact-govuk-one-login/index.njk
+++ b/src/components/contact-govuk-one-login/index.njk
@@ -77,7 +77,6 @@
 {# START page content #}
 
 {% block pageContent %}
-
 <h1 class="govuk-heading-xl">{{'pages.contact.header' | translate}}</h1>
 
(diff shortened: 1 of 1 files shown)


alex9smith:
Thanks for the detailed PR description - makes my job a lot easier! 

I've checked this out and run it locally but it's not quite working as I'd expect. You can see in the recording below that when I use the keyboard to navigate to the 'Use webchat' link and press enter the webchat window pops up for an instant then minimises again. Navigating back to that link while the chat window is open closes the window (with the right focus 👏🏻 ) but then does the same thing when trying to open the window again. 

Any ideas? It's possible this is unrelated to your change and if so we should tackle it separately. 

https://github.com/govuk-one-login/di-account-management-frontend/assets/6362602/32f1b7f4-b153-47e1-b1e7-3fc209e6ef19

danacotoran:
Thanks @alex9smith!
Ah that is strange. It was certainly working for me locally before I opened the PR. Let me give it a look and see why that might be happening.

danacotoran:
@alex9smith I worked out what was causing it, but I have no clue why: in a separate commit, before I opened the PR, I added the webchat-adjacent scripts to the `scripts` block at the end of the `body` tag since that's generally considered to be good practice. 
I thought it would be a harmless change, but for some reason it broke the behaviour of the "Use webchat" button! 
I'm not sure why yet, I'll look into it later, but I guess the important thing is that it should work now 😅  Sorry about that!

PR #1245, opened 2024-03-28
OLH-1670: Remove 'Find a grant' from services list

alex9smith:
### What changed

The 'Find and apply for a grant' service was wrongly in both the list of account services and standalone services. This meant that if a user had the service card they'd see it under both 'Your accounts' and 'Other services you've used'.

### Why did it change

We missed this duplication when we changed these lists in #1192. I've checked all the other IDs in the lists and there's no other duplicate entries so this is the only service we need to check.

## Testing
I've run this locally and given myself a card for this service. I can see it now only appears once in the correct section:

localhost_6001_your-services (1)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -192,8 +192,6 @@ export const getAllowedServiceListClientIDs: string[] = [
   "mortgageDeed",
   VETERANS_CARD_PROD,
   VETERANS_CARD_NON_PROD,
-  FIND_AND_APPLY_FOR_A_GRANT_PROD,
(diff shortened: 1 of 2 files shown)


saralk (approved):
Is it the case that something shouldn't be in the allowed account list and allowed service list at the same time?

If so, it could be worth adding some kind of validator that runs in the github actions

alex9smith:
Hah, yes I had the same thought while I was on my lunch break. I've added a unit(?) test that checks there's no overlap as there shouldn't be unless we have a very special case (and then we'll want some tests anyway).

PR #1256, opened 2024-04-03
OLH-1678 - forbidden error.

andrew-moores:
## Proposed changes

Add a handler for CSRF failures which have been observed in production.
### What changed
A new handler to deal with CSRF validation failures.  When a CSRF validation occurs the error is caught by the new handler, logged and the user is re-directed back to their Your Services page.  There are a number of integration tests that check this behaviour and they been altered to use a new test method that describes the expected behaviour when CSRF validation fails.
The order that error handlers were added in the main app file was inconsistent with the docs so I re-ordered them to remove any future confusion.

### Why did it change
The new handler will remove the 403 error that was being detected by DynaTrace.
The existing integration tests were found to have an error in them that prevented individual scenarios being run.

## Testing

I have tested locally.  The changes should be deployed to dev and tested there as they impact most routes.  The _csrf hidden form field can be easily changed using browser dev tools to simulate a failed validation.

If, after testing, we decide that redirecting to the Your Services page is not the correct behaviour we can change this behaviour simply by updating `src/handlers/csrf-error-handler.ts` and `test/utils/behaviours.ts` to support testing.

### DIFF
--- a/src/components/change-email/tests/change-email-integration.test.ts
+++ b/src/components/change-email/tests/change-email-integration.test.ts
@@ -70,9 +70,9 @@ describe("Integration:: change email", () => {
     app = await require("../../../app").createApp();
     baseApi = process.env.AM_API_BASE_URL;
 
-    request(app)
+    await request(app)
       .get(PATH_DATA.CHANGE_EMAIL.url)
-      .end((err, res) => {
+      .then((res) => {
         const $ = cheerio.load(res.text);
         token = $("[name=_csrf]").val();
         cookies = res.headers["set-cookie"];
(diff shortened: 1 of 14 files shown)


alex9smith (approved):
Nice work on sorting out the tests here

src/components/change-email/tests/change-email-integration.test.ts
```diff
@@ -92,14 +92,15 @@ describe("Integration:: change email", () => {
     request(app).get(PATH_DATA.CHANGE_EMAIL.url).expect(200, done);
   });
 
-  it("should return error when csrf not present", (done) => {
-    request(app)
+  it("should redirect to you services when csrf not present", async () => {
```
peterfajemisincabinetoffice, line 95:
```suggestion
  it("should redirect to your services when csrf not present", async () => {
```

peterfajemisincabinetoffice (approved):

PR #1258, opened 2024-04-03
OLH-1662 add CTA if the user only has one type of MFA

saralk:
### What changed

This adds a CTA to encourage the user to set up a secondary backup method for 2FA.

### Why did it change

To encourage adoption of backup methods

- [X] Application configuration is up-to-date
- [X] Added to deployment steps
- [X] Added to local startup config

## Testing

Tested Locally:

Screenshot 2024-04-03 at 16 38 18

Screenshot 2024-04-03 at 16 38 33


src/components/security/security-controller.ts
```diff
@@ -59,11 +59,21 @@ export async function securityGet(req: Request, res: Response): Promise<void> {
       })
     : [];
 
+  const hasAuthAppMethod = req.session.mfaMethods?.some((method) => {
+    return method.mfaMethodType === "AUTH_APP";
+  });
```
saralk, line 64: Do we need this? I added this in, but based on the wording of the message, I think we would always want to show the message when the user only has one method

src/locales/en/translation.json
saralk, line 1: we need welsh translations

src/components/security/security-controller.ts
peterfajemisincabinetoffice: @saralk , do you still need an answer to this?

saralk: https://gds.slack.com/archives/C02F70D2LQN/p1712149445312499

peterfajemisincabinetoffice (approved):

PR #1263, opened 2024-04-04
OLH-1664 Add another MFA page.

saralk:
### What changed

Adds a new route (behind a feature flag), which allows the user to select what kind of backup 2fa method they are going to add.

Currently, once selected, the user gets an error. We will not launch this until the entire flow is ready.

Also includes a small refactor, we used to have a partial template for each password change request type. I have refactored this as the only difference was the translation keys.

Screenshot 2024-04-08 at 08 50 17
Screenshot 2024-04-08 at 08 50 27
Screenshot 2024-04-08 at 08 59 22


src/components/add-mfa-method/add-mfa-methods-controller.ts
```diff
@@ -0,0 +1,29 @@
+import { Request, Response, NextFunction } from "express";
+import { PATH_DATA } from "../../app.constants";
+
+export function addMfaMethodGet(req: Request, res: Response): void {
+  const helpText = `<p>${req.t("pages.addMfaMethod.app.help.text1")}</p><p>${req.t("pages.addMfaMethod.app.help.text2")}</p>`;
```
andrew-moores, line 5: Isn't this usually done in the nunjucks template?

src/app.ts
```diff
@@ -21,6 +21,7 @@ import {
   getSessionExpiry,
   getSessionSecret,
   supportActivityLog,
+  supportMfaUpsell,
```
andrew-moores, line 24: What is the relevance of "Upsell" in the name?

src/components/add-mfa-method/add-mfa-methods-controller.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export function addMfaMethodPost(
+  req: Request,
+  res: Response,
+  next: NextFunction
+): void {
+  const { addMfaMethod } = req.body;
+
+  if (addMfaMethod === "sms") {
```
andrew-moores, line 16: Can "sms" be a constant?  (Same for "app" below).

src/components/add-mfa-method/tests/add-mfa-methods-controller.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      redirect: sandbox.fake(),
+    };
+    next = sinon.spy();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it("should return 400 status code when addMfaMethod is 'sms'", () => {
```
andrew-moores, line 29: Could the name of the test describe what happens from the users perspective?  e.g. should not allow user to add a backup MFA method of APP when the primary method is SMS.

src/locales/en/translation.json
```diff
@@ -181,6 +181,23 @@
...
+      },
+      "app": {
+        "title": "Authenticator app for smartphone, tablet or computer",
+        "help": {
+          "title": "What is an authenticator app?",
+          "text1": "An authenticator app creates a security code that helps confirm it's you when you sign in.",
+          "text2": "You can use an authenticator app on your smartphone, tablet or desktop computer. Download an authenticator app for your smartphone or tablet from your app store or search online for an authenticator app for your computer."
+        }
+      },
+      "backLink": "I do not want to add another way to get security codes",
```
andrew-moores, line 198: Is the design correct that this doesn't end with a full stop?

saralk: good point, i will clarify that

src/components/add-mfa-method/add-mfa-methods-controller.ts
saralk: Done, and refactored a bit to use the constants more widely

src/components/add-mfa-method/tests/add-mfa-methods-controller.test.ts
saralk: agreed and changed

andrew-moores (approved):

PR #1277, opened 2024-04-10
OLH-1665 Build the new 'Set up an authenticator app' page

saralk:
### What changed

Creates the flow for adding an authenticator app TOTP to an account. 

Screen recording of the flow:

https://github.com/govuk-one-login/di-account-management-frontend/assets/748653/db562d09-9bb3-4d75-bf3e-d78fb0ea6882


package.json
```diff
@@ -83,10 +83,12 @@
     "libphonenumber-js": "^1.10.60",
     "nunjucks": "^3.2.4",
     "openid-client": "^5.6.5",
-    "pino": "^8.20.0",
+    "otplib": "^12.0.1",
+    "pino": "^8.19.0",
     "pino-http": "^9.0.0",
     "prettier": "^3.2.5",
-    "uuid": "^9.0.1",
```
aenciso, line 89: Can't we use uuid 9 instead?

saralk: yes, i'm not sure how it got downgraded.

peterfajemisincabinetoffice (approved):

src/components/enter-password/_change-phone-number.njk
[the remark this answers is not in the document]
saralk: Originally there were four sub-templates, one for each type of enter password method. Each template was identical, except for the translation strings.
So instead of creating a fifth file for the MFA method, I moved the content of the sub-templates into the main one, and get the translations in the controller.

PR #1298, opened 2024-04-15
OLH-1709: Add helper function for TxMA header

alex9smith:
### What changed

Add a helper function to get the value of the TxMA device information header.

I considered extending the set-local-vars middleware to put this value in `res.locals` instead, but it's possible that in the future we'll send audit events in a context that doesn't have clean access to the `Response` object. I thought it'd be a nicer pattern to call this function in the controller and pass the result down to the service that's making the API call or sending the audit event.

There's also a tiny amount of logic here, so a separate function is easier to unit test.

This function isn't called anywhere yet - we have other tickets to deal with including the value in our audit events.

### Why did it change

 We'll need to use this value everywhere we send audit events so we want to have one place we deal with fetching it.

### DIFF
--- a/src/utils/txma-header.ts
+++ b/src/utils/txma-header.ts
@@ -1,0 +1,5 @@
+import { Request } from "express";
+
+export function getTxmaHeader(req: Request): string {
+  return (req.headers["txma-audit-encoded"] as string) ?? "";
+}
--- a/src/utils/txma-header.ts
+++ b/src/utils/txma-header.ts
@@ -1,11 +1,10 @@
 import { Request } from "express";
 import { logger } from "./logger";
 
-export function getTxmaHeader(req: Request, trace: string): string {
+export function getTxmaHeader(req: Request, trace: string): string | undefined {
   if (req.headers["txma-audit-encoded"]) {
     return req.headers["txma-audit-encoded"] as string;
   } else {
     logger.warn({ trace: trace }, "Missing Txma-Audit-Encoded header");
-    return "";
   }
 }
--- a/src/utils/txma-header.ts
+++ b/src/utils/txma-header.ts
@@ -1,5 +1,11 @@
 import { Request } from "express";
+import { logger } from "./logger";
 
-export function getTxmaHeader(req: Request): string {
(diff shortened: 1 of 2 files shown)


src/utils/txma-header.ts
```diff
@@ -0,0 +1,9 @@
+import { Request } from "express";
+
+export function getTxmaHeader(req: Request): string {
```
aenciso, line 3: I know this is short enough but it could be simplified to
`req.headers["txma-audit-encoded"] as string ?? "";`

aenciso (approved):

src/utils/txma-header.ts
```diff
@@ -0,0 +1,5 @@
...
+  return (req.headers["txma-audit-encoded"] as string) ?? "";
```
andrew-moores, line 4: If the header isn't present rather than returning an empty string, which could also indicate other issues such as a coding error, would it be better to use a special case value?  If it is missing is it something we'd want to instrument and alert on to flag as a problem?

alex9smith: I think it'd be sensible to send a special case value but that needs to be the same value across the programme and I've not seen any mention of doing so in the docs for this work. I've asked in the [initiative Slack channel](https://gds.slack.com/archives/C06GXH2TNSW/p1713255873299489) but we may not get a consensus quickly. 

You're definitely right about monitoring - we should at least have a `WARN` level message if it's not present. I'll add that now.

alex9smith: I got a reply 

>  the preference will be not to send the property and any value if you don't have one for it.  This would be consistent with what the cloudfront function being used to generate the audit header does it if can't find a value.  This will help consumers to not need to content check the field as to whether the value is genuine or the missing placeholder text.

We'd want to handle that behaviour at the point we generate the audit events so do you think an empty string is ok here as a 'falsy' value, or would it be better to return `undefined`?

andrew-moores: Sounds good.  If we don't have a special value then I feel like undefined would be a slightly better option.

andrew-moores (approved):

PR #1305, opened 2024-04-16
OLH-1710 Add TxMA header to contact us audit event

saralk:
### What changed

Add to audit events, the TxMA encoded data that is added to request headers.

### Why did it change

To support the new TxMA audit event changes.

### Related links

Uses the library created here: https://github.com/govuk-one-login/di-account-management-frontend/pull/1298


src/services/event-service.test.ts
```diff
@@ -80,6 +81,9 @@ describe("eventService", () => {
         "2024-09-12T00:00:00.000Z"
       );
       expect(result.timestamp).to.equal(1726099200);
+      expect(result.restricted.device_information.encoded).to.equal(
+        "test-txma-header"
```
andrew-moores, line 85: Should we use a b64 encoded string of the length we are expecting here to make it more realistic?

saralk: agreed, i've done this now

aenciso (approved):

PR #1309, opened 2024-04-18
OLH-1713 Send TxMA header in all calls to Auth's APIs

saralk:
### What changed

- Update the `getRequestConfig` function to use named parameters rather than positional. This makes it easier to manager, since there are a lot of parameters. We don't need to use `null` anymore
- Use the `getTxmaHeader` utility function to populate the field, and pass on the txma header
- Update tests to check the above
- Update types to support the above.

### Why did it change

These headers need to be passed through in our API calls, to support TxMA's new fraud detection system


src/components/delete-account/delete-account-service.ts
```diff
@@ -16,21 +16,22 @@ export function deleteAccountService(
...
   ): Promise<boolean> {
     const { status } = await axios.client.post<void>(
       API_ENDPOINTS.DELETE_ACCOUNT,
       {
         email: email,
       },
-      getRequestConfig(
+      getRequestConfig({
         token,
-        null,
```
aenciso, line 28: If this is removed then sourceIp would become validationStatuses parameter, is this correct?

aenciso: No worries, just noticed that getRequestConfig has been refactored down here. All good.

src/utils/http.ts
```diff
@@ -28,25 +28,35 @@ export function createApiResponse(
...
+  txmaAuditEncoded?: string;
+}): AxiosRequestConfig {
   const config: AxiosRequestConfig = {
     headers: {
       Authorization: `Bearer ${token}`,
     },
     proxy: false,
   };
 
-  if (validationStatues) {
```
aenciso, line 47: Good catch

PR #1314, opened 2024-04-19
OLH-1724 Delete account and backchannel logout journeys don't delete user's sessions

andrew-moores:
## Proposed changes

When a user deletes their account or logs out via the global logout also delete the Express session.

### What changed

The destroyUserSessions function and its callers.

### Why did it change

The destroyUserSessions function in the session-store module attempts to delete all entries in the user sessions table.  However, it did not destroy the Express session which resulted in the session persisting after the user thought they were logged out.  Adding a final call to delete the Express session ensured everything was removed.  The session-store destroyUserSessions function was refactored to highlight the two distinct actions that it takes, firstly to find all the users records in the session store and delete them and secondly to destroy the Express session.

### DIFF
--- a/src/utils/session-store.ts
+++ b/src/utils/session-store.ts
@@ -1,3 +1,4 @@
+import { Request } from "express";
 import {
   DynamoDBClient,
   QueryCommand,
@@ -9,6 +10,7 @@ import { getDBConfig } from "../config/aws";
 import { logger } from "./logger";
 import connect_dynamodb from "connect-dynamodb";
 import { Store } from "express-session";
+import { ERROR_MESSAGES } from "../app.constants";
 
 // the value of the USER_IDENTIFIER_IDX_ATTRIBUTE must match the indexed attribute in SessionsDynamoDB table
 // defined in `../../deploy/template.yaml`.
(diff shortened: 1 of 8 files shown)


andrew-moores:
@alex9smith do we want to consider deleting all the sessions from Dynamo in the logout-controller as well?

alex9smith:
Short answer - yes we should do that as well. We should log them out of all active sessions across any devices they're using.

In practice it probably doesn't matter because when someone clicks the logout button:
1. We delete their express session & DynamoDB record for that session
2. We redirect the user to Auth's 'logged out' page
3. Auth sends us a [backchannel logout notification](https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/log-your-users-out/#responding-to-logout-notifications-from-gov-uk-one-login) to the `/global-logout` route 
4. That controller will run and delete all the rest of the user's sessions

But I think it's better to be explicit in our intentions on the logout controller too. It'll be easier for future us to debug if it's clear from 5 lines of code that a user clicking log out will end all their active sessions.

src/utils/session-store.ts
```diff
@@ -58,26 +60,47 @@ async function getSessions(subjectId: string): Promise<string[]> {
...
+      } out of ${results.length} failed`
+    );
+  }
+}
+
+async function deleteExpressSession(req: Request) {
+  req.session.destroy((err) => {
+    if (err) {
+      logger.error(
+        { trace: "hello" },
```
alex9smith, line 85: Looks like you left this debug trace in

alex9smith (approved):
Good to go once you've taken that debug trace out

peterfajemisincabinetoffice (approved):

PR #1347, opened 2024-05-01
OLH-1721 Add error logging when we don't have a translation for content

aenciso:
### What changed

Added an error log for missing Welsh translation. Fallback language had to be removed from the initial i18next configuration in order to explicitly throw an error and still return the English translation. There are other alternatives like using `missingKeyHandler` option but it would not allow us to do what we wanted so a safeTranslation function was added in instead.

### Why did it change

To find in our logs which translation keys we are missing

## Testing

Unit tests
Local testing

## How to review

- Change a key in `cy/translation.json` e.g.  pages.security.header to pages.security.headerX
- Run the application locally and go to http://localhost:6001/security?lng=cy
- The error message should be displayed in the console as  ERROR (di-account-management-frontend/32301): TranslationError: key 'pages.security.header' missing for requested language.


src/utils/safeTranslate.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const safeTranslate = (
+  translate: TFunction<"translation", undefined>,
+  key: string,
+  options?: Record<string, any>
+): translateResult => {
+  const result: translateResult = translate(key, options);
+  if (result === key) {
+    logger.error(
+      `TranslationError: key '${key}' missing for requested language.`
```
alex9smith, line 15: It's not clear what's in the options object - do we have easy access to which language we're looking for? It'd be helpful to know if the missing key is English or Welsh. 

It's not going to be hard to search for the key in two files, but if we can easily put it in the logs we should

aenciso: Done

package.json
```diff
@@ -23,7 +23,7 @@
     "pretty": "prettier --write \"*/**/*.{ts,json}\"",
     "start": "node dist/server.js",
     "test:audit": "npm audit",
-    "test:integration": "rm -rf test/coverage && NODE_ENV=development nyc mocha -r dotenv/config -P tsconfig.mocha.json  \"src/**/*-integration.test.ts\"",
+    "test:integration": "rm -rf test/coverage && NODE_ENV=test nyc mocha -r dotenv/config -P tsconfig.mocha.json  \"src/**/*-integration.test.ts\"",
```
peterfajemisincabinetoffice, line 26: Is this change necessary?

peterfajemisincabinetoffice (approved):

PR #1385, opened 2024-05-20
OLH-1745: Update content on update phone journey

danacotoran:
### What changed

Update some of the content on the update phone number journey (and a few bits in some other journeys) in light of bringing account management frontend more in line with auth. 

The content updates include:
-  moving the "Problems with the code" dropdown under the submit button on all versions of this page
- changing "I do not want to change my [thing]" to "Cancel and go back to Security"
- updating how we play back the new phone number in the update phone number journey:  changed the language and the way the phone number is displayed. Now we consistently display the last 4 digits, in bold, instead of the obfuscated "*******1111" format
- use uppercase "Security" across the journey when talking about the Security page 

### Why did it change
More consistent language across the old and new journeys.

## Testing
Mel has checked this on dev 

## How to review
Double check the content matches the figma and that all the welsh content is there. 
Check that nothing's broken as a consequence of the removal of the old phone number obfuscation util func.

### DIFF
--- a/src/components/change-phone-number/index.njk
+++ b/src/components/change-phone-number/index.njk
@@ -6,52 +6,51 @@
 {% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
 {% set pageTitleName = 'pages.changePhoneNumber.title' | translate %}
 {% block content %}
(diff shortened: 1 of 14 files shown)


alex9smith (approved):
👍🏻

PR #1390, opened 2024-05-22
OLH-1794 - Ensure controllers and services can make an update auth app call to the API

peterfajemisincabinetoffice:
## Proposed changes
OLH-1794 - Ensure controllers and services can make an update auth app call to the API
https://govukverify.atlassian.net/browse/OLH-1794
### What changed

New controller and services implementation for updating the authenticator app as part of account management using the new method management API

### Why did it change
So support users updating their registered authenticator aop

### Related links
https://govukverify.atlassian.net/browse/OLH-1794

## Testing

Deploy to dev environment, attempt to update app and ensure successful and a call to method management update endpoint is called


src/components/change-authenticator-app/index.njk
```diff
@@ -0,0 +1,68 @@
...
+
+                {{ govukButton({
+                "text": "general.continue.label" | translate,
+                "type": "Submit",
+                "preventDoubleClick": true
+                }) }}
+            </form>
+        </li>
+    </ol>
+{% endblock pageContent %}
```
danacotoran, line 68: this can just be `{% endblock %}`

danacotoran (commented):
@peterfajemisincabinetoffice thank you for your work on this!

There are loads of changes and I haven't been able to dive deep into everything you've added as part of this PR so this will not be a super comprehensive review, sorry! 

Some questions popped to mind as I was going through your PR and I thought I would submit a review anyway before I go away for the long weekend, even though it's not the most detailed one!

I was just wondering about the new template you are adding `change-authenticator-app/index.njk`, as well as more generally, the new logic around generating the QR code, validating MFA and so on. 
A lot of it seems to be similar to the work that has already been done on the "add MFA method" piece I believe Saral worked on.
I was wondering if it might be possible to reuse some of the code from the "add MFA method app" piece as the templating/content here seems quite similar to what was implemented there (if not the same), and much of the logic in the controller seems to be the same as well. 

In one case you're setting up an entirely new mfa method and in the other you're replacing an existing mfa method with another one but there seems to be significant repetition/overlap between the two bits of work and I was wondering if it could be more DRY friendly.

I might be totally off base here and there might be some clear reasons against doing what I'm proposing. If that is the case, I'd be interested in learning more about it!

src/components/change-authenticator-app/index.njk
peterfajemisincabinetoffice: Thank Dana, this has been changed.

peterfajemisincabinetoffice:
> @peterfajemisincabinetoffice thank you for your work on this!
> 
> There are loads of changes and I haven't been able to dive deep into everything you've added as part of this PR so this will not be a super comprehensive review, sorry!
> 
> Some questions popped to mind as I was going through your PR and I thought I would submit a review anyway before I go away for the long weekend, even though it's not the most detailed one!
> 
> I was just wondering about the new template you are adding `change-authenticator-app/index.njk`, as well as more generally, the new logic around generating the QR code, validating MFA and so on. A lot of it seems to be similar to the work that has already been done on the "add MFA method" piece I believe Saral worked on. I was wondering if it might be possible to reuse some of the code from the "add MFA method app" piece as the templating/content here seems quite similar to what was implemented there (if not the same), and much of the logic in the controller seems to be the same as well.
> 
> In one case you're setting up an entirely new mfa method and in the other you're replacing an existing mfa method with another one but there seems to be significant repetition/overlap between the two bits of work and I was wondering if it could be more DRY friendly.
> 
> I might be totally off base here and there might be some clear reasons against doing what I'm proposing. If that is the case, I'd be interested in learning more about it!

Thanks Dana for looking at this in detail. 
(1) The decision to have separate controllers for the add vs. update mfa was discussed with @alex9smith , however I can see there is definitely room for refactoring for re-use of code within controllers them selves and I will get that done.

(2) As per the template(change-authenticator-app/index.njk), logic for generating QR, etc. we are still undecided if this should be done by Auth's API or how it is done currently, see: https://govukverify.atlassian.net/browse/OLH-1824 So probably best leaving that bit until we are clear.

peterfajemisincabinetoffice:
> > @peterfajemisincabinetoffice thank you for your work on this!
> > There are loads of changes and I haven't been able to dive deep into everything you've added as part of this PR so this will not be a super comprehensive review, sorry!
> > Some questions popped to mind as I was going through your PR and I thought I would submit a review anyway before I go away for the long weekend, even though it's not the most detailed one!
> > I was just wondering about the new template you are adding `change-authenticator-app/index.njk`, as well as more generally, the new logic around generating the QR code, validating MFA and so on. A lot of it seems to be similar to the work that has already been done on the "add MFA method" piece I believe Saral worked on. I was wondering if it might be possible to reuse some of the code from the "add MFA method app" piece as the templating/content here seems quite similar to what was implemented there (if not the same), and much of the logic in the controller seems to be the same as well.
> > In one case you're setting up an entirely new mfa method and in the other you're replacing an existing mfa method with another one but there seems to be significant repetition/overlap between the two bits of work and I was wondering if it could be more DRY friendly.
> > I might be totally off base here and there might be some clear reasons against doing what I'm proposing. If that is the case, I'd be interested in learning more about it!
> 
> Thanks Dana for looking at this in detail. (1) The decision to have separate controllers for the add vs. update mfa was discussed with @alex9smith and we decided to go with separate controllers, routes etc, however I can see there is definitely room for refactoring for re-use of code within controllers them selves and I will get that done.
> 
> (2) As per the template(change-authenticator-app/index.njk), logic for generating QR, etc. we are still undecided if this should be done by Auth's API or how it is done currently, see: https://govukverify.atlassian.net/browse/OLH-1824 So probably best leaving that bit until we are clear.

Thanks @danacotoran , this has also been addressed. Please help to re-review

alex9smith (approved):
Nice one Peter - I think that's a good refactor we've ended up with.

danacotoran (approved):
Thanks Peter 👍

PR #1435, opened 2024-06-13
OLH-1823 Build the new pages to add a phone number as a backup method

saralk:
### What changed

- Add a new screen for the user to enter their phone number
- Update the "add mfa method" screen to only show one option is the user already has an MFA App method enabled.

Update the "check your phone" screen to include the concept of intents...

We already have a check your phone flow (where the user enters the 6 digit code to confirm their number), this is used during the "change your number" flow

This PR updates it to include the concept of "intents"

We are able to re-use the flow here for adding a new MFA method, when we show the page to the user, we include an `intent` param in the URL, which can either be `changePhoneNumber` or `addMfaMethod`.

Then when the number is confirmed, we can do a different action.

### Why did it change

To support the "add backup mehtod" functioanloity

## How to review

I have made changes to an existing flow (check your phone), so please look at that code in depth


src/components/add-mfa-method-sms/add-mfa-method-sms-controller.ts
```diff
@@ -0,0 +1,48 @@
...
+  res: Response
+): Promise<void> {
+  res.render("add-mfa-method-sms/index.njk");
+}
+
+export async function addMfaSmsMethodPost(
+  req: Request,
+  res: Response
+): Promise<void> {
+  //TODO do something with this
```
peterfajemisincabinetoffice, line 20: Did you mean to leave this here?

src/components/add-mfa-method-sms/add-mfa-method-sms-routes.ts
```diff
@@ -0,0 +1,34 @@
...
+);
+
+router.post(
+  PATH_DATA.ADD_MFA_METHOD_SMS.url,
+  requiresAuthMiddleware,
+  validateStateMiddleware,
+  addMfaSmsMethodPost
+);
+
+router.get(
```
peterfajemisincabinetoffice, line 27: FYI We generally configure all the confirmation routes here: https://github.com/govuk-one-login/di-account-management-frontend/blob/main/src/components/update-confirmation/update-confirmation-routes.ts

peterfajemisincabinetoffice (approved):

PR #1469, opened 2024-06-25
OLH-1825 Update the journeys we've already done to reflect changes in the API spec

saralk:
### What changed

Update our use of the account management API, to reflect the changes that have been made to it.

### Why did it change

So our app still works with the new API

### Related links

https://github.com/govuk-one-login/account-management-stubs/pull/167
https://github.com/govuk-one-login/openapi-specs/pull/15


peterfajemisincabinetoffice (approved):

peterfajemisincabinetoffice (changes requested):
You have conflicting files. plese resolve and then I will approve

src/components/security/security-controller.ts
```diff
@@ -24,8 +24,8 @@ export async function securityGet(req: Request, res: Response): Promise<void> {
       ? req.session.mfaMethods.map((method) => {
```
aenciso, line 24: Perhaps it is worth renaming `method` to `mfaMethod` so the call is mfaMethod.method rather than method.method which is slightly confusing or just deconstruct it at top level so the calls don't need to be refactored

saralk: agreed

peterfajemisincabinetoffice (approved):

PR #1487, opened 2024-07-01
OLH-1923: Update analytics package to v2.0.1

alex9smith:
### What changed

Update the analytics package to v2.0.1. 

In this PR we go from v1.0.3 to v2.0.1, which includes the major version update to v2. This introduced a new parameter (see the V2 upgrade [in the README](https://www.npmjs.com/package/@govuk-one-login/frontend-analytics)) that allows for form field tracking if we're sure that there's no personal data in our forms. We can't make that assumption as we have phone numbers, email addresses etc. so I've set the new param to `isDataSensitive: true`. 

### Why did it change

This is required before we can switch off the old UA implementation. 

### Related links

https://gds.slack.com/archives/C05JLGJU6BV/p1719825552253399

## Testing

I've run this locally against the test GA4 property and seen the events firing.

### DIFF
--- a/src/components/common/layout/base.njk
+++ b/src/components/common/layout/base.njk
@@ -150,7 +150,8 @@
       }, {
         disableGa4Tracking: {{isGa4Disabled}},
         disableUaTracking: {{isUaDisabled}},
-        cookieDomain: "{{analyticsCookieDomain}}"
+        cookieDomain: "{{analyticsCookieDomain}}",
+        isDataSensitive: true
       });
     }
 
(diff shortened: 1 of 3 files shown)


aenciso (approved):

src/components/common/layout/base.njk
```diff
@@ -150,7 +150,8 @@
       }, {
         disableGa4Tracking: {{isGa4Disabled}},
         disableUaTracking: {{isUaDisabled}},
-        cookieDomain: "{{analyticsCookieDomain}}"
+        cookieDomain: "{{analyticsCookieDomain}}",
+        isDataSensitive: true
```
danacotoran, line 154: My understanding is that `isDataSensitive` should be `true` by default now, and that the way to disable it (I'm really not sure why we should ever want to disable it – the instructions on the README haven't really expanded on that) is to set `isDataSensitive: false`

alex9smith: It is `true` by default but I thought it'd be better here to be explicit and set it ourselves since it's a setting that protects personal data

PR #1518, opened 2024-07-15
Bau revert to node 20

alex9smith:
### What changed

Revert to the latest version of Node 20 which is the current LTS release and tell Dependabot to ignore major version upgrades. We'll still get PRs for minor and patch updates to the image but this means we'll stay on Node 20 until we intentionally upgrade.

### Why did it change

This is for 2 reasons:

1. Our runtime version policy[^1] says we "must be on the most recent LTS version of a given runtime"
2. We're having issues with Dynatrace on our frontend and previously we fixed similar problems by downgrading Node to a version with longer Dynatrace support.

[^1]: https://team-manual.account.gov.uk/development-standards-processes/policies-and-standards/runtime-version-policy/#runtime-version-policy

## Testing

I've built the image locally and verified it runs the frontend.


aenciso:
I think .nvmrc also needs to be updated

alex9smith:
I can see the update to `.nvmrc` [here](https://github.com/govuk-one-login/di-account-management-frontend/pull/1518/files#diff-2c8effdf840690ccb88c7e21e56148978c6adb2c6926530c58ff0e47a9588b44R1) - does that not show up for you @aenciso?

aenciso:
Weird, it wasn't showing. All good.

.github/dependabot.yml
```diff
@@ -37,3 +37,7 @@ updates:
       - dependencies
     commit-message:
       prefix: BAU
+    ignore:
+      - dependency-name: "*"
```
aenciso, line 41: Should not this be just node?

aenciso (approved):

PR #1547, opened 2024-07-24
OLH-1488 Update 404 page

aenciso:
### What changed

Changed the way 404 page behaves depending on whether a user is logged in or not.
For logged in users, navigation is displayed and the main CTA button goes to `your-services` page
For not logged in users, navigation is not shown and the main CTA button goes to sign in page.

### Why did it change

Because that’s the more appropriate place to send users who get a “not found” while navigating their One Login home, and less confusing when we don't show the navigation

- [ ] Added a local variable `res.locals.isUserLoggedIn`

## Testing

Local testing for both logged in and not logged in users

image
image

## How to review

Navigate to any non-existent page to see the 404 page. Check both navigation and CTA  button copy

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -69,6 +69,7 @@ import { safeTranslate } from "./utils/safeTranslate";
 import { addMfaMethodSmsRouter } from "./components/add-mfa-method-sms/add-mfa-method-sms-routes";
 import { deleteMfaMethodRouter } from "./components/delete-mfa-method/delete-mfa-method-routes";
 import { changeDefaultMethodRouter } from "./components/change-default-method/change-default-method-routes";
+import { isUserLoggedInMiddleware } from "./middleware/is-user-logged-in-middleware";
 
 const APP_VIEWS = [
   path.join(__dirname, "components"),
@@ -167,6 +168,7 @@ async function createApp(): Promise<express.Application> {
   app.use(startRouter);
   app.use(logoutRouter);
   app.use(enterPasswordRouter);
+  app.use(isUserLoggedInMiddleware);
   app.use(changeEmailRouter);
   app.use(updateConfirmationRouter);
   app.use(changePasswordRouter);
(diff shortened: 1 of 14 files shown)


danacotoran (commented):
Hi @latifcabinetoffice and/or @aenciso thank you for picking this up 🙌 

I've checked the 404 page and it displays as intended, whether the user is logged in or not.
A couple bits I thought would be worth mentioning:
- Firstly: I didn't notice a UCD sign-off in the ticket so I was just going to say, If the new content hasn't been double-checked by a UCD person yet, would you mind arranging a review? 
-  Secondly: the update in the way the secondary nav active item is set means that "Security" no longer appears as "active" on the security sub-journeys (change email/password/delete account etc). I think it'd have to be set explicitly on all those journeys now

aenciso:
@danacotoran on the second point, you're right I completely missed those scenarios. The reason being is that when a user is signed in, the navigation were showing security as an active item even in 404 and 500 error pages which I think is misleading. What we can do is to either set each page manually or create a general function that sets it active depending on the URL. I'm more inclined for the latter but we can discuss it later if you have time.

peterfajemisincabinetoffice (approved):

danacotoran:
Just commenting to say, for our future selves can we please add some more detail to the commit message and/or PR description as "Update 404 page"  doesn't quite cover everything that has changed in this and why it has changed.
For instance, why did the nunjucks `translate` filter have to be updated in `src/config/nunjucks.ts`? Would that immediately be obvious to a developer debugging a potential problem with this code in the future?

aenciso:
@danacotoran sorry, yes, all commits were separate but when this PR was approved I squashed them all before hitting merge. Also remember that in this repo we squash our commits when we merge so those commit messages are lost unfortunately.
The nunjucks change is to handle a TranslationError that's been cluttering our logs.

PR #1597, opened 2024-08-12
OLH-2014 Remove default selected option for add backup method page

aenciso:


danacotoran (commented):
Hi, @aenciso many thanks for picking this up!

Would you be able to check with Mel that the content (for the error message as well as the heading text on this page) is correct please? 
As the content in the screenshot on the JIRA ticket is different from the content that we have on here currently which I believe might be different from the latest version of the content in Figma.

saralk (approved):

PR #1599, opened 2024-08-13
OLH-2052 Add new DWP integration client service card and update existing one

peterfajemisincabinetoffice:
## Proposed changes
OLH-2052 Add new DWP integration client service card and update existing one
https://govukverify.atlassian.net/browse/OLH-2052

### What changed
Add new entry and update URL for existing entry in translation files for DWP to test new RP in integration environment only. 

### Why did it change

So that DWP can test their new service cards before releasing to production.

## Testing

As it is integration environment only, there is no point deploying this to dev environment to check. i confirmed with Ben this is OK and matches what was done before by @alex9smith  as this is not the same as adding a new RP to prod.

## How to review
Code completeness and invalid JSON.

### DIFF
--- a/src/locales/cy/translation.json
+++ b/src/locales/cy/translation.json
@@ -863,9 +863,15 @@
       },
       "sqae3L7gOdizeRqFMw_KCDlhcyg": {
         "header": "Repay and manage benefit money you owe",
-        "description": "Ad-dalu a rheoli arian budd-dal sy’n ddyledus gennych.",
+        "description": "Ad-dalu a rheoli arian budd-dal sy'n ddyledus gennych.",
         "link_text": "Go to your repay and manage benefit money you owe account",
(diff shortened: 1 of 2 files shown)


saralk (approved):

src/locales/cy/translation.json
```diff
@@ -863,9 +863,15 @@
       },
       "sqae3L7gOdizeRqFMw_KCDlhcyg": {
         "header": "Repay and manage benefit money you owe",
-        "description": "Ad-dalu a rheoli arian budd-dal sy’n ddyledus gennych.",
+        "description": "Ad-dalu a rheoli arian budd-dal sy'n ddyledus gennych.",
+        "link_text": "Go to your repay and manage benefit money you owe account",
```
danacotoran, line 867: Just to confirm, do we not have the Welsh language translation yet? 
Or is it not relevant at this stage and can be added later?

peterfajemisincabinetoffice: Thanks Dana, yes we don't have the full welsh translations. I would think its not relevant as they are currently testing

PR #1625, opened 2024-08-20
OLH-2053-endpoint-will-be-changed-to-phone-number-in-the-method-management-api

latifcabinetoffice:
## Proposed changes

[OLH-2053] Replaced mfamethod scheme to have SMS phone number instead of endpoint
### What changed

1) mfa schema updated to differentiate between AUTH APP and SMS
2) replaced endpoint with phone number
3) Fixed tests
4) Replaced logic which added backup method as AUTH APP by default 

### Why did it change

In line with mfa api

### Permissions

- [ ] This PR adds or changes permissions

## Testing

Deployed updated stub and frontend to dev and changed phone number and added a new SMS mfa backup. 

[OLH-2053]: https://govukverify.atlassian.net/browse/OLH-2053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/check-your-phone/check-your-phone-controller.ts
+++ b/src/components/check-your-phone/check-your-phone-controller.ts
@@ -153,24 +153,22 @@ async function handleAddMfaMethod(
     return false;
   }
 
-  if (defaultMfaMethod.method.mfaMethodType !== "SMS") {
-    return false;
-  }
-
   try {
-    defaultMfaMethod.method.phoneNumber = newPhoneNumber;
-    updateInput.credential = "no-credentials";
-    updateInput.mfaMethod = {
-      ...defaultMfaMethod,
-      mfaIdentifier: defaultMfaMethod.mfaIdentifier + 1,
-      priorityIdentifier: "BACKUP",
-      method: {
-        mfaMethodType: defaultMfaMethod.method.mfaMethodType,
-        phoneNumber: newPhoneNumber,
-      },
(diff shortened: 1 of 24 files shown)


src/components/check-your-phone/check-your-phone-controller.ts
```diff
@@ -140,31 +140,32 @@ async function handleAddMfaMethod(
   service: CheckYourPhoneServiceInterface,
   trace: string
 ): Promise<boolean> {
-  const smsMFAMethod = req.session.mfaMethods.find(
+  const defaulMfaMethod = req.session.mfaMethods.find(
     (mfa) => mfa.priorityIdentifier === "DEFAULT"
```
saralk, line 144:
```suggestion
    (mfa) => mfa.priorityIdentifier === "DEFAULT" && mfa.method.mfaMethodType === "SMS"
```

If I'm undestanding the logic correctly here, this would mean we could get rid of the if statement on line 154?

latifcabinetoffice: annoyingly get type error's with this change, without the if statement, it believe's the mfa method could still be auth app and doesn't allow for the use of phoneNumber

peterfajemisincabinetoffice (approved):

PR #1681, opened 2024-09-16
OLH-2126 UCD Changes to translation strings

saralk:
## Proposed changes

Update some translation strings based on feedback from UCD


src/components/security/index.njk
```diff
@@ -17,7 +17,7 @@
         items: [
           {
             href: enterPasswordUrl + "&type=changeEmail&edit=true",
-            text: 'general.change' | translate,
+            text: 'pages.security.accountDetails.summaryList.changeEmail' | translate,
```
danacotoran, line 20: Thanks for making these changes @saralk.
Agree with the proposition to have longer link text (as long as UCD are happy with the layout implications).  

The summary list component [accepts a `visuallyHiddenText` parameter](https://design-system.service.gov.uk/components/summary-list/#:~:text=To%20give%20more%20context%2C%20add%20visually%20hidden%20text%20to%20the%20links.%20This%20means%20a%20screen%20reader%20user%20will%20hear%20the%20row%20action%20and%20the%20%E2%80%98key%E2%80%99%20label%20for%20the%20information%20it%20will%20affect%2C%20like%20%E2%80%98Change%20name%E2%80%99%20or%20%E2%80%98Change%20date%20of%20birth%E2%80%99.) for accessibility purposes – so assistive tech users can understand what the purpose of row actions is out of context.
Here we've changed the link text to be more descriptive but we haven't altered or removed the visually hidden text associated with the links. This means there will be repetition for assistive tech users. 

Screenshot 2024-09-17 at 09 31 07

Can we remove the visually hidden text from the row actions please?

danacotoran (changes requested):

src/components/security/index.njk
saralk: thanks @danacotoran i have made that change

danacotoran (changes requested):
Hi @saralk, thanks for the updates. 
A couple more points, sorry:
- since this update touches on the designs that are currently live in production (I mean specifically the updates to the "Change password/email" links), it'd be good if it could be signed off by UCD. I wasn't sure whether it had been as it wasn't mentioned on the ticket or on the PR description
- do we need to add the Welsh language versions for the new link text (Change password and Change email) as well? 

Thanks!

danacotoran (approved):

PR #1688, opened 2024-09-17
OLH-2134 UCD changes for the add backup method page

saralk:
- Change URL to /choose-backup
- Update some strings
- Add hint text


src/app.constants.ts
```diff
@@ -13,27 +13,27 @@ export const PATH_DATA: Record<
...
   },
   ADD_MFA_METHOD_SMS: {
-    url: "/add-mfa-method-sms",
+    url: "/choose-backup-sms",
     event: EventType.VerifyCodeSent,
     type: UserJourney.AddMfaMethod,
   },
   ADD_MFA_METHOD_SMS_CONFIRMATION: {
-    url: "/add-mfa-method-sms-confirmation",
+    url: "/choose-backup-sms-confirmation",
```
danacotoran, line 36: I don't think any URLs should be changing as part of this ticket, except for the `choose-backup` one.

src/components/add-mfa-method/index.njk
```diff
@@ -11,12 +11,16 @@
...
+    hint: { 
+        text: 'pages.addMfaMethod.sms.hint' | translate
+    }
 } %}
 {% set appItem = {
     value: 'app',
     text: 'pages.addMfaMethod.app.title' | translate,
-    hint: 'pages.addMfaMethod.app.hint' | translate
+    hint: {
+        text: 'pages.addMfaMethod.app.hint' | translate
```
danacotoran, line 22: These is no value for this in the translation file. As per the design in figma, it doesn't look like there should be a hint for the auth app option, so I think the `hint` option can be removed here.

danacotoran (changes requested):
Hi @saralk, thanks for making these changes.

I've left some inline comments, I hope they make sense to you.

An additional point I wanted to make is that I think it'd be best for the sake of consistency that we also rename the files/directories/translation keys/anything else associated with the `choose-backup` screen. 

Thanks!

danacotoran (commented):
Thanks for this @saralk, looks good to me! so I'm happy to approve once the commit history has been tidied up and the minor typo I noticed has been fixed.

danacotoran (approved):

PR #1845, opened 2024-12-02
OLH-2027 Throw errors in frontend for critical pathways

peterfajemisincabinetoffice:
## Proposed changes

OLH-2027 Throw errors in frontend for critical pathways

### What changed

Add a global try/catch block to wrap all page load 

### Why did it change
So that if error occurs during page loads, relevant metrics and alarms are triggered.

### Testing

Deploy to dev and test


src/components/enter-password/enter-password-routes.ts
```diff
@@ -4,25 +4,25 @@ import {
...
+  globalTryCatch(enterPasswordGet)
 );
 
 router.post(
   PATH_DATA.ENTER_PASSWORD.url,
   requiresAuthMiddleware,
   validateEnterPasswordRequest(),
   refreshTokenMiddleware(),
-  asyncHandler(enterPasswordPost())
+  globalTryCatch(enterPasswordPost())
```
latifcabinetoffice, line 25: should this be async?

peterfajemisincabinetoffice: Thanks for spotting, this has been changed to use correct global try/catch and async handler

latifcabinetoffice (approved):

PR #1858, opened 2024-12-09
OLH-2163 - update analytics package to 3.0.1

peterfajemisincabinetoffice:
## Proposed changes
Redeploy - OLH-2163 - update analytics package to 3.0.1

### What changed
Bump the frontend analytics package to version 3.0.1. 
Followed the documentation outlined here 
https://govukverify.atlassian.net/wiki/spaces/DIFC/pages/4416143374/Frontend+Analytics+v3+Implementation+Guide 
In essence:
- the global flag has been refactored: originally it was written as `disableGa4Tracking` which was confusing to read as the negative meant the Boolean value against the flag would be mirrored (setting it to `true` would turn the package off). The default value for the global flag is now set to `false`
- tracker specific feature flags have been added (I couldn't see any reason to turn any of them off per se, so I hard coded them to `true` as I imagine as long as GA4 is on, these trackers should all be on as well 🤔  open to opposing perspectives)
- the "select tracking" flag is off in production while PA team test and confirm it's working as expected
- `edit=true` parameter has been added to all the links leading into the "change" type journeys in account management

### Related links
https://govukverify.atlassian.net/wiki/spaces/DIFC/pages/3801317710/Frontend+Analytics+Implementation+Guide

## How to review
Baseline manual test with a browser extension (like GA debug) to ensure that everything is still working as before.
[Slack conversation](https://gds.slack.com/archives/C012GEZBZM0/p1729690434085639) covers most of the points here. 
Check nothing is missing as per the implementation guide.

### DIFF
--- a/src/components/security/index.njk
+++ b/src/components/security/index.njk
@@ -16,7 +16,7 @@
       actions: {
         items: [
           {
-            href: enterPasswordUrl + "&type=changeEmail&edit=true",
+            href: enterPasswordUrl + "&type=changeEmail",
             text: 'pages.security.accountDetails.summaryList.changeEmail' | translate
           }
         ]
@@ -33,7 +33,7 @@
       actions: {
         items: [
           {
-            href: enterPasswordUrl + "&type=changePassword&edit=true",
+            href: enterPasswordUrl + "&type=changePassword",
             text: 'pages.security.accountDetails.summaryList.changePassword' | translate
           }
         ]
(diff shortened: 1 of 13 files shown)


src/components/security/index.njk
```diff
@@ -16,7 +16,7 @@
       actions: {
         items: [
           {
-            href: enterPasswordUrl + "&type=changeEmail&edit=true",
+            href: enterPasswordUrl + "&type=changeEmail",
```
saralk, line 19: is this correct? Do we not need the edit=true param here (it is the only place in this PR where we are removing it)

latifcabinetoffice (approved):

src/components/security/index.njk
peterfajemisincabinetoffice: @saralk, going by this PR where it was removed: https://github.com/govuk-one-login/di-account-management-frontend/pull/1800/files#diff-54b690956706838b48a5df100ac823f3816560712e4728fa673083d47899accf. Essentially what we are doing is re-instating the changes in this PR

PR #2000, opened 2025-02-28
OLH-2460 - Update frontend to use library for HMRC check

peterfajemisincabinetoffice:
## Proposed changes

OLH-2460 - Update frontend to use library for HMRC check

### What changed

HMRC clientId are loaded from clientRegistry by calling the new filterClients method and passing the isHmrc=true

### Why did it change

So that we can start using clientRegistry in frontend.

### Related links

https://govukverify.atlassian.net/browse/OLH-2460


src/config.ts
```diff
@@ -315,7 +317,16 @@ export const getAllowedAccountListClientIDs: string[] = [
   RURAL_PAYMENT_WALES_NON_PROD,
 ];
 
-export const hmrcClientIds: string[] = [HMRC_NON_PROD, "hmrc"];
```
saralk, line 318: i think we don't need to recreate the array here...

In this file, for example: https://github.com/govuk-one-login/di-account-management-frontend/blob/main/src/utils/yourServices.ts#L100

We could have something like...

```
const hasDetailedCard = getClient(getAppEnv(), service.client_id).isHmrc
```

saralk (approved):

PR #2038, opened 2025-03-19
OLH-2519 Change the way that translations are loaded

saralk:
### What changed

If the feature flag is enabled for using the client library, then we should load the translation files directly and make the translation bundle, rather than using the i18next filesystem backend.

### Why did it change

addResourceBundle overrides the translations rather than merge them, so welsh translations weren't appearing.

### Testing

### Sign-offs


latifcabinetoffice:
@saralk ran locally and for some reason some parts are translating but not others:
image

latifcabinetoffice (approved):

saralk:
@latifcabinetoffice I believe that those strings have not been translated yet, as we have not gone live with MFA management yet

PR #2041, opened 2025-03-21
[OLH-2517] Copy asset script now cleans dist/public/scripts repository of old files before copying over current files

latifcabinetoffice:
## Proposed changes

[OLH-2517] Copy asset script now cleans dist/public/scripts repository of old files before copying over current files

### What changed

Addition of rm -r /dist/public/scripts and recursive copy of anything within src/public/scripts

### Why did it change

Previously the builds were copying over new changes but did not delete file's that were removed. This led to an inconsistent state during local development. Additionally, hardcoding file's in the copy asset script had potential to cause error's and prevent any files copying over.

## How to review

Tested js file's are still loaded in local development and deleted scripts and checked /dist/public/scripts within the docker container to validate if it was actually removed. Deployed to dev and validated all scripts that should be present are present.

[OLH-2517]: https://govukverify.atlassian.net/browse/OLH-2517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alex9smith (commented):
Right, and this is `copy-scripts` which is called by `copy-assets` which in turn is called by `npm run build`. So that hopefully will fix the issue. Nice!

alex9smith (approved):

alex9smith:
(sorry, I forgot to click approve with my comment above)

PR #2044, opened 2025-03-21
[DO NOT MERGE]: OLH 2526: activity history tweaks

danacotoran:
**DO NOT MERGE**: the link to the reporting form is yet to be confirmed, and Welsh translations are missing for the new content. 
These changes do not switch activity history on. That is being done as part of [this PR](https://github.com/govuk-one-login/di-account-management-frontend/pull/2043).
### What changed
[OLH-2526: Reword "RSA allow list" to "Activity Log Allow List" in the code](https://github.com/govuk-one-login/di-account-management-frontend/commit/7bfc02113d8b906bb37be5ecfee3cdac57f07d4c) 

The original plan was that activity log and the (OLH-owned) RSA journey would go hand in hand – there would be an allow list of services whose users would all have the ability to access the RSA journey. All users who could access the activity history, could also access our RSA journey. 

In the meanwhile the requirement has shifted, so there will be a rollout of the activity log with a "new" RSA capability (in the shape of a form developed by the auth team). The "old" RSA journeys (developed by OLH) are to remain switched off for now.

There is now a requirement to be able to toggle these three things on/off independently (the activity history view, the "old" OLH RSA journeys and the link to the "new" RSA form).

In light of this, the terminology around "RSA allow list" has been updated in our code to reflect the updated requirement/what is happening now: the activity log is essentially being rolled out with a RSA-lite (which may be turned on/off depending on capacity, tbc) where instead of each item in the activity history having its own Report link, there is a catch-all contact link at the top of the page.
The allowlist-ing is of services whose users get to access the activity history(with or without RSA).

[OLH-2526: Update reporting info paragraph](https://github.com/govuk-one-login/di-account-management-frontend/commit/9042ab17e72658f4201914e9fa8ffd8a041c5892) 

Update content on the reporting info paragraph at the top of the activity history page. 
There has been a decision to introduce a separate flag to allow us to switch the ability to navigate to the auth RSA form on/off where needed (i.e. when support requests exceed current capacity)

When the flag is on, the content will appear as a dropdown with a link to the form, and a paragraph telling users to change their password.
When the flag is off, the content will appear as one paragraph advising users to change their password.

### Why did it change
Activity history due to be rolled out to approx 10% of OLH users on the 27th of March. 

### Testing

### Sign-offs


.env.sample
```diff
@@ -19,6 +19,7 @@ SERVICE_STORE_TABLE_NAME=user_services
 SUPPORT_LANGUAGE_CY=1
 SESSION_STORE_TABLE_NAME=account-mgmt-frontend-SessionStore
 SUPPORT_ACTIVITY_LOG=1
+REPORTING_FORM_ENABLED: "1"
```
latifcabinetoffice, line 22: Should we keep the same semantics, e.g. SUPPORT_REPORTING_FORM

src/components/activity-history/index.njk
```diff
@@ -10,7 +10,7 @@
 {% endif %}
 {% set securityNoticeHtml %}
   <p class="govuk-body">{{ 'pages.activityHistory.securityNoticeContent1' | translate | safe | replace("[changePasswordLink]", changePasswordLink) }}</p>
-  <p class="govuk-body">{{ 'pages.activityHistory.securityNoticeContent2' | translate | safe | replace("[reportActivityLink]", contactLink) }}</p>
+  <p class="govuk-body" data-test-id="content-for-reporting-form-enabled">{{ 'pages.activityHistory.securityNoticeContent2' | translate | safe | replace("[reportActivityLink]", contactLink) }}</p>
```
latifcabinetoffice, line 13: should this be data-test-id

.env.sample
danacotoran: Good point, happy to change it!

src/components/activity-history/index.njk
danacotoran: I used `data-test-id` as it seems to be a pre-existing convention for selectors meant specifically for test purposes. But it seems like there are a variety of selectors used to verify existence of element(s) in the markup, so the convention isn't always followed.
Maybe something we can all have a chat about at some point!

latifcabinetoffice (approved):

PR #2053, opened 2025-03-25
OLH-2140: Add a backlink to the "Set up an authenticator app" page

alex9smith:
### What changed

Add a backlink to the "Set up an authenticator app" page which takes the user back to the enter password page with the correct `type` parameter for this journey. This works for both the GET and POST responses when the template is
re-rendered with errors.

I'm 90% sure this works fine with our state machine. I've clicked back and forth a few times and I'm able to complete the journey. We'll be able to catch any errors in our next team review before we put this live.

To do this, I've added another optional parameter to `renderMfaMethodPage`. Two optional parameters is ok for this function, but  think if we add another we should consider exposing the parameters as a separate interface object instead.

image

### Why did it change

To match the Figma design.


alex9smith:
I'm closing this as we realised we don't want the backlink on this page - it was a mistake in the design. The logic for adding one is still useful though. I've pulled that commit out into a separate PR in #2058.

PR #2068, opened 2025-03-31
OLH-2534: Set up an API client class for the method management API

alex9smith:
### What changed

Set up a new API client class and helper types that we can use to interact with the method management API. This class isn't used anywhere at the moment. In future PRs I plan to work through the places where the existing functions are called and replace those with this new client.

My hope is to replace calls like this:

```typescript
await updateMfaMethod({
  email: string;
  credential?: string;
  updatedValue?: string;
  otp: string;
  mfaMethod?: MfaMethod;
}, {{
  accessToken: string;
  sourceIp: string;
  sessionId: string;
  persistentSessionId: string;
  userLanguage: string;
  clientSessionId: string;
  txmaAuditEncoded: string;
}}) 
```

with something like: 
```typescript
mfaClient = new MfaClient(res.locals.user.publicSubjectId, getRequestConfig({...});

const response = await mfaClient.update(newMfaMethod);
```

### Why did it change

I started out reviewing the API calls against [the published spec](https://github.com/govuk-one-login/openapi-specs/blob/main/auth/account-management-api/account-management-with-mfa-method-management.yaml) but I found the current code really hard to follow and understand. We've got big data objects with the `any` type, raw request objects being passed several layers down, loads of abstraction layers etc. Peter and I had already found one area where we've made a mistake in our request body and I'd like to set up the type system to help us avoid future similar bugs. 

Much of the data that's being passed around is request scoped and we can encapsulate it in a class that we instantiate in the controller. We can then simplify all the actual API calls to just the information we care about for the body of that request. 

I suspect we ended up with the current code as a result of how we structured the work. We built this journey by journey and didn't consider the bigger picture of how we should work with the API. I take responsibility for that as I was the one who planned this work way back when.

## How to review

Please check the requests / responses against [the api spec](https://github.com/govuk-one-login/openapi-specs/blob/main/auth/account-management-api/account-management-with-mfa-method-management.yaml) and check you're happy with the level of test coverage.


latifcabinetoffice:
Very nice changes! What was the error in the request config / & has that been resolved?

alex9smith:
> What was the error in the request config

We were trying to change the MFA identifier when we were updating a method rather than passing the same ID back with updated values. Yes Peter has already sorted that in the work he was doing.

PR #2089, opened 2025-04-07
OLH-2548: Remove the SUPPORT_LANGUAGE_CY feature flag

liegeandlief:
### What changed

* Removed `SUPPORT_LANGUAGE_CY` feature flag and defaulted any relying code to true
* Set `REPORT_SUSPICIOUS_ACTIVITY` to "1" in `.env.sample` to ensure that unit tests pass when the repo has just been set up

### Why did it change

We will always support Welsh in the future so the feature flag is no longer required.

### Related links

https://govukverify.atlassian.net/browse/OLH-2548


alex9smith (approved):
🚀

PR #2106, opened 2025-04-10
[OLH-2260] show different text about webchat on the contact page when …

liegeandlief:
### What changed

Show alternative content about webchat on contact page when JavaScript is disabled

### Why did it change

So that users with JavaScript understand why webchat isn't available to them and what they can do about ut.

### Related links

https://govukverify.atlassian.net/browse/OLH-2260

### Sign-offs


alex9smith (commented):
Feel free to tell me "no, that's too hard", but is it possible to add an integration test for this behaviour?

alex9smith (approved):
Makes sense, let's do that.

src/assets/scss/application.scss
```diff
@@ -316,6 +316,14 @@ $nav-button-padding-bottom: $nav-link-padding + $nav-link-outline-thickness;
   @include govuk-font($size: 48, $weight: bold);
 }
 
+.contact-webchat-js-enabled {
+  display: none;
+}
+
+.js-enabled .contact-webchat-js-enabled {
+  display: unset;
```
danacotoran, line 324: I think this should be okay but need to double check browser usage stats to confirm that we definitely don't need to support IE11 anymore (or any browser where unset could be problematic).

src/components/contact-govuk-one-login/index.njk
```diff
@@ -116,15 +116,51 @@
...
+            <p class="govuk-body">{{ paragraph }}</p>
+          {% endfor %}
+
+          {% set listLinks =  'pages.contact.section3.webchat.javascriptDisabled.listLinks' | translate({ returnObjects: true }) %}
 
-        <p class="govuk-body"><button type="button" class="launch-webchat-link" data-launch-webchat hidden>{{ 'pages.contact.section3.webchat.linkText' | translate }}</button></p>
+          {% if listLinks.length %}
+            <ul class="govuk-list govuk-list--bullet">
+              {% for link in listLinks %}
+                <li><a class="govuk-link" href="{{link["href"]}}" target="_blank" rel="noreferrer noopener">{{link["text"]}}</a></li>
```
danacotoran, line 159: [Design system guidance](https://design-system.service.gov.uk/styles/links/#opening-links-in-a-new-tab) suggests that we should avoid opening links in a new tab without making it clear to users that this is the expected behaviour. 
It might be good to bring this up with the UCD team and see what their thoughts are! 

If this behaviour is something UCD requested, is there any reason why "opens in a new tab" couldn't be added into the content?

liegeandlief:
> Makes sense, let's do that.

Ticket created: https://govukverify.atlassian.net/browse/OLH-2582

danacotoran (commented):
Thanks for picking this up!
Just had a couple thoughts/questions, the main one being regarding a UCD decision.

src/assets/scss/application.scss
liegeandlief: @danacotoran where can I go to check which browsers we support?

PR #2109, opened 2025-04-11
[OLH-1983] serve valid robots.txt to hopefully show crown logo in search engine results

liegeandlief:
### What changed

- Updates `robots.txt` to ensure its format is valid and to allow access to all URLs in production and no URLs otherwise.
- Removes unnecessary meta robots tag from contact page.

### Why did it change

When investigating why Google is not displaying the favicon in search results for the contact page I noticed that there were some issues with the way we are controlling robot access to the site.

Standard practice is to allow access to all URLs and then blacklist specific URLs rather than to block all URLs and then whitelist certain URLs.

These current changes allow access to all URLs with nothing blacklisted. This also means that the favicons are accessible and so hopefully Google will start to show them in the search results for the contact page (I'll make a ticket to check in a few weeks).

**If there are any specific URLs we don't want to show in search results then please comment on this PR and I'll add disallow rules for them.**

Having checked the docs at https://developers.google.com/search/docs/appearance/favicon-in-search our icons appear to be configured correctly to show in the search results so hopefully these changes will be enough to get it working.

### Related links

https://govukverify.atlassian.net/browse/OLH-1983

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1389,10 +1389,9 @@ Resources:
         - Type: fixed-response
           FixedResponseConfig:
             ContentType: text/plain
-            MessageBody: >
-              "user-agent: *"
-              "disallow: /"
-              "allow: /contact-gov-uk-one-login"
+            MessageBody: |
+              User-agent: *
+              Disallow:
             StatusCode: 200
       Conditions:
         - Field: path-pattern
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1382,24 +1382,24 @@ Resources:
       FromPort: !Ref ContainerPort
       ToPort: !Ref ContainerPort
 
-  # ApplicationLoadBalancerListenerRuleRobots:
-  #   Type: AWS::ElasticLoadBalancingV2::ListenerRule
-  #   Properties:
-  #     Actions:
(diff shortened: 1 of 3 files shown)


alex9smith (changes requested):
Ah yes of course. Google can't show the favicon if we've blocked them 🤦🏻 

I think allowing all pages to be crawled in production is ok - most of the app is behind the sign in anyway so we won't be exposing any data. We can identity the Google crawl bot by the useragent so let's make a note on our followup ticket to also check crawl volumes to double check they're not causing us any issues (I don't expect there will be any). 

What we do need to change here is to keep blocking all URLs in all our other environments. I reckon the simplest way to do that is with a [Cloudformation condition](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html) to pick between this listener rule in production and a new one that blocks everything for everywhere else. We already have an [`IsProduction` condition](https://github.com/govuk-one-login/di-account-management-frontend/blob/main/deploy/template.yaml#L114) that you can use.

alex9smith (approved):

liegeandlief:
@alex9smith where would I need to go to check crawl volumes?

PR #2110, opened 2025-04-11
[OLH-2266] migrate post-deploy tests from puppeteer to playwright

liegeandlief:
### What changed

Post-deploy tests migrated from Puppeteer to Playwright.

### Why did it change

- Playwright offers a better experience installing and managing browsers and does not to be told where the executable is for each browser. This is useful because running tests locally on `arm64` machines requires running against the locally installed version of Chrome. With Playwright this will just work without having to update variables to point to the executable.
- Having worked with both Puppeteer and Playwright in the past Playwright has always been much less flaky and has more extensive cross-browser capabilities which may be useful in future tests.
- In the last couple of years Playwright seems to have replaced Puppeteer as the go-to tool.

### Related links

https://govukverify.atlassian.net/browse/OLH-2266

### Sign-offs

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,15 +115,26 @@ docker exec -it account-management-frontend /bin/sh
 
 ### Post-deploy tests
 
-The post-deploy tests are written with Jest, [jest-cucumber](https://www.npmjs.com/package/jest-cucumber) and [Puppeteer](https://pptr.dev/).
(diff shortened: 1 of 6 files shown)


alex9smith (approved):
Nice! I'm on an Intel (AMD64) Mac still so I could test both running in Docker and against my local Chrome.

PR #2116, opened 2025-04-15
OLH-2534: Use the MFA API client when retrieving users MFA methods

alex9smith:
### What changed

Use the MFA API client when retrieving users MFA methods and update the session type definition to use the `MfaMethod` type exported by the API client. 

This meant I could remove all the ugly `as unknown as MfaMethod` type casts I needed before.

It's safe to do this now because there's sufficient overlap between the two types since we updated the old type to use a string as the identifier[^1].

[^1]: https://github.com/govuk-one-login/di-account-management-frontend/pull/2052

### Why did it change

This is the next step in updating the frontend to use the MFA API client.

### Testing

I've deployed this to dev and verified it still picks up the user's MFA methods


liegeandlief (approved):
Lots of code removed, great stuff!

alex9smith:
Code out, tests in 👍🏻

PR #2119, opened 2025-04-15
OLH-2534: Fix the API calls to create an MFA method

alex9smith:
### What changed

This PR includes 2 fixes to the MFA API client:
1. Pass the OTP in the API call when a user creates an SMS method.
2. Use the correct payload structure when creating a new method.

We only need to pass the OTP when a user is creating an SMS method. When they create an authenticator app our frontend can (and does) do all the required validation on the OTP they provide. 

The journey to create a new SMS method isn't migrated to use the API client yet - I'll include the OTP in the calls from the controller when I do that. This PR just has the API client changes.

### Why did it change

Andrew and I found these mistakes when we were debugging the journeys in staging. 

### Related links

[API spec](https://github.com/govuk-one-login/openapi-specs/blob/main/auth/account-management-api/account-management-with-mfa-method-management.yaml)

### DIFF
--- a/src/utils/mfaClient/index.ts
+++ b/src/utils/mfaClient/index.ts
@@ -10,9 +10,11 @@ import {
   AuthAppMethod,
   MfaClientInterface,
   MfaMethod,
+  CreateMfaPayload,
 } from "./types";
 import { HTTP_STATUS_CODES } from "../../app.constants";
 import { getTxmaHeader } from "../txma-header";
+import { validateCreate } from "./validate";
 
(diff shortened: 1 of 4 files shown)


liegeandlief:
@alex9smith SonarCloud has some complaints. I think you just need to change `throw` to throw()` maybe

liegeandlief (changes requested):
@alex9smith SonarCloud has some complaints. I think you just need to change throw to throw()` maybe

liegeandlief (approved):

PR #2137, opened 2025-04-22
OLH-2534: Use the MFA API client when a user updates their SMS methods

alex9smith:
### What changed

Use the MFA API client in all remaining journeys that go through the 'check your phone' controller. These are:
- Update your phone number on an existing method
- Switch your default method from auth app -> SMS

There's a couple of things to note in these changes. I'll put a summary here but please see the individual commit messages for the details and the context of code changes.

1. We can now pass an SMS OTP to the client when we call `.update()`. This is required when changing a phone number or switching from auth app to SMS, but not when we make other changes eg. switching backup to primary.
2. The API spec doesn't currently include the OTP in the update call, but I've spoken to Auth and they've confirmed this is an oversight and that we should include it in the same way as when we create a new SMS method. 
3. I think there's a bug in the logic we had to find a user's MFA method when they're updating the phone number on an existing method. I've created a separate ticket to look into that some more as the scope of my current ticket is already huge.
4. My change to switching your backup method won't work in staging at the moment as the API doesn't currently support this operation. It will in the future though, so I've just made our frontend changes to be the expected future state. 

### Why did it change

We're replacing all the calls to the MFA API to use the new API client. 

### Sign-offs
Not required

### DIFF
--- a/src/components/check-your-phone/check-your-phone-controller.ts
+++ b/src/components/check-your-phone/check-your-phone-controller.ts
@@ -24,6 +24,11 @@ import {
 } from "../check-your-email/types";
 import { logger } from "../../utils/logger";
 import { createMfaClient, formatErrorMessage } from "../../utils/mfaClient";
+import {
+  ApiResponse,
+  MfaClientInterface,
+  MfaMethod,
+} from "../../utils/mfaClient/types";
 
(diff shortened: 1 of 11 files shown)


liegeandlief (approved):
Looks good to me and looks like you've removed more than you've added which is always nice! 🙌

PR #2141, opened 2025-04-23
Olh 2142 ucd fixes remove a back up mfa journey 7

latifcabinetoffice:
## Proposed changes

[OLH-2142] UCD fixes for remove backup mfa journey 7
### What changed

changed type from removeMfaMethod to removeBackup
added backlink on confirmation page
changed url from remove-method to remove-backup

### Why did it change

UCD review to align with designs


src/components/delete-mfa-method/index.njk
```diff
@@ -21,5 +22,7 @@
         "type": "Submit",
         "preventDoubleClick": true
         }) }}
+    
+        <p class="govuk-body"> <a href="/manage-your-account" class="govuk-link" rel="noreferrer noopener">{{cancelText}}</a></p>
```
liegeandlief, line 26: suggestion: use `"SECURITY" | getPath` to prevent hardcoding the path

liegeandlief (changes requested):
Just one small change but otherwise looks good to me

liegeandlief:
Also looks like some tests are failing @latifcabinetoffice

src/components/delete-mfa-method/index.njk
latifcabinetoffice: ah apols it looks like when I created the draft it auto assigned one-login-home, @alex9smith is it possible to disable auto assign if the pr is in draft?

liegeandlief: weird it doesn't usually auto-assign when the PR is a draft

PR #2143, opened 2025-04-23
BAU: Do linting validation on the Cloudformation template

alex9smith:
### What changed

There's a new(ish) option to `sam validate` which calls out to `cfn-lint` to actually check the structure of the template and all the resources, rather than the basic checks you get without the option.

### Why did it change

We want to catch any errors early, before we deploy, and this is another tool to help us do that.

### Testing
I've run the command locally against `main` to check it passes:

```bash
frontend (BAU-sam-validate-with-lint) 4f376eac $ sam validate -t deploy/template.yaml --lint
/Users/alexsmith/di/frontend/deploy/template.yaml is a valid SAM Template
```

### DIFF
--- a/.github/workflows/validate-sam-template-dev.yml
+++ b/.github/workflows/validate-sam-template-dev.yml
@@ -48,4 +48,4 @@ jobs:
           aws-region: eu-west-2
 
       - name: SAM Validate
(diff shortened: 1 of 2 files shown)


liegeandlief (approved):
Nice! Extra linting's always good

PR #2145, opened 2025-04-24
OLH-2534: Remove the last of the old MFA code

alex9smith:
### What changed

There's 3 main changes in this PR:
1. Use the MFA API client when changing your default auth app
2. Use the MFA API client when changing your default SMS method to an auth app
3. Remove all of the now unused MFA code

### Why did it change

This is (hopefully) the final step in replacing all calls to the MFA API with our API client.

### Testing

I've run this locally and stepped through the journeys, checked the validation errors still appear etc. Once this is merged I'll do the same in staging.

## How to review

It's probably easiest to review this commit by commit - the journey changes are contained in the relevant commit and the rest is mostly deleting things.


liegeandlief:
@alex9smith in the description you say that these changes include `Use the MFA API client when changing your backup auth app to a new one`. As far as I can tell it shouldn't be possible to change the backup auth app, only the default auth app. At least from a user's perspective this seems to be true.

Does the client implement this functionality because auth support it in their backend? If so is it in the client but unused?

alex9smith:
Ah sorry. My description isn't quite right there. This is either:
1. Changing a **default** SMS method to an auth app method
2. Updating a **default** auth app method with a new credential

As you say, the only operations the user's allowed to do on a backup method is promote it to default or delete it.

PR #2149, opened 2025-04-24
OLH-2143: MFA journey 8 UCD fixes (part 1)

liegeandlief:
### What changed

(sorry for the many changes in one PR, it's hard to untangle them into discrete chucks when working through the MFA UCD fixes, and fixing one things leads to something else breaking which in turn needs fixing 😬)

- Add back link to the screen for entering phone number when changing the default MFA method
- Use the shared template for entering phone number when entering phone number when changing the default MFA method
- Ensure the check your phone screen renders with the same options on both GET and POST requests
- Correct the back link URL when going back from the check your phone screen to the screen for entering phone number when changing the default MFA method
- Correct the ‘use a different phone number’ link in the the three different MFA phone number scenarios so that it goes back to the expected screen in each case
- Added ‘Cancel and go back to Security’ link to the resend phone code screen
- Added back link to the resend phone code screen
- Ensure the resend phone code screen renders with the same options on both GET and POST requests

### Why did it change

To fix issues found in UCD testing of the MFA flows. Also fixes other issues which were found whilst fixing the UCD identified issues.

### Related links

https://govukverify.atlassian.net/browse/OLH-2143


alex9smith (approved):
🚀 There might be some conflicts with my PR https://github.com/govuk-one-login/di-account-management-frontend/pull/2145. I've switched off auto-merge there and will resolve them on Monday morning.

PR #2157, opened 2025-04-25
a user can only have a default or backup method. On the frontend the …

latifcabinetoffice:
…user can only change their default sms method phone number. When a user has a default SMS and a backup SMS it was just the first one to return that get's updated.

### What changed

A user can only have a default or backup method. On the frontend the user can only change their default sms method phone number. When a user has a default SMS and a backup SMS it was just the first one to return that get's updated. This ensure's that it is the default SMS method that gets updated. If a user wants to change their backup SMS they will have to remove it and then add it again. This has been raised with Ben and Sam.

### Why did it change

Potential to update backup SMS method when trying to change default phone number.

### DIFF
--- a/src/components/check-your-phone/check-your-phone-controller.ts
+++ b/src/components/check-your-phone/check-your-phone-controller.ts
@@ -114,7 +114,9 @@ async function changePhoneNumberwithMfaApi(
 
     case INTENT_CHANGE_PHONE_NUMBER: {
       const smsMFAMethod = currentMfaMethods.find(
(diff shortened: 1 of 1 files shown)


liegeandlief (approved):

alex9smith:
Looks good. You'll probably have a conflict with https://github.com/govuk-one-login/di-account-management-frontend/pull/2145 now I've merged it, but it should be obvious where to move your change to. Give me a shout if it's not.

PR #2168, opened 2025-04-30
[OLH-2590] MFA error and validation improvements

liegeandlief:
### What changed

- show an error message when an invalid OTP code is sent to backend API
- remove duplication of MFA code validation logic
- move password validation logic into handler to ensure that renders of the page all have the same options
- move phone OTP validation logic into handler to ensure that renders of the page all have the same options
- execute phone number validation in handlers to ensure that renders of the page all have the same options

### Why did it change

To improve the experience for the user and to tidy up the code.

### Related links

https://govukverify.atlassian.net/browse/OLH-2590


alex9smith (approved):
Nice work. This is much easier to follow now everything's using that auth app page handler Same for moving the phone code validation back out into the controllers. Plus it looks like we have some test coverage on that behaviour now! 

Is the 'wrong OTP' error the only one from the API that we can reasonably handle at this point?

liegeandlief:
> Nice work. This is much easier to follow now everything's using that auth app page handler Same for moving the phone code validation back out into the controllers. Plus it looks like we have some test coverage on that behaviour now!
> 
> Is the 'wrong OTP' error the only one from the API that we can reasonably handle at this point?

Yes it's the only one we can reasonably handle. Most of the errors should never actually happen because we've already performed the necessary checks on the frontend to ensure we're making a valid request to the backend. Therefore if they do then something's wrong in our code and letting the error bubble up seems like the sensible thing to do.

PR #2178, opened 2025-05-08
OLH-2608: Update MFA client to not reject on 400 errors

liegeandlief:
### What changed

Updated the MFA client config so that requests don't reject on 200, 204 and 400 errors.

### Why did it change

We expect the MFA backend API to respond with 400 errors when bad requests are made from the frontend. We want to be able to handle these responses and show error messages to the user rather than an exception being thrown.

### Related links

https://govukverify.atlassian.net/browse/OLH-2608


alex9smith (approved):
Nice find. I had no idea axios rejected the promise on a status code greater than 300. I thought it was only if the request failed for some reason like a timeout.

I think this explains a lot of the puzzling I'd done about not seeing error logs the last few weeks as well.

PR #2180, opened 2025-05-09
Disallow everything except specific paths in robots.txt

liegeandlief:
### What changed

Rather than allowing everything in robots.txt it has been changed to only allow crawling of the contact page and things in the `public` and `assets` folders.

### Why did it change

To try and get Google to display nicer search results for the contact page, which is the only page we want to be crawled and indexed.

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1409,7 +1409,10 @@ Resources:
               - IsProduction
               - |
                 User-agent: *
-                Disallow:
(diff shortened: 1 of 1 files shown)


TaherKapasi (approved):

danacotoran:
I hope this works 🤞

liegeandlief:
> I hope this works 🤞

Me too although I'm not holding out hope!

PR #2197, opened 2025-05-20
OLH-2659: Set the device intelligence domain

alex9smith:
### What changed

Add a new environment variable to control this and set it to `account.gov.uk` in all environments except local.

All variables in `res.locals` are automatically available in the templates.

### Why did it change

This prevents issues with the request header size being too large in pre-production environments ([see Slack](https://gds.slack.com/?redir=%2Farchives%2FC02SM1N3P2S%2Fp1746790187823409%3Fname%3DC02SM1N3P2S%26perma%3D1746790187823409)).

- [x] Added to local startup config

### Testing

I've run this locally and can see the cookie is still being set.


deploy/template.yaml
```diff
@@ -201,6 +201,8 @@ Mappings:
       MISSIONLABWEBSOCKETADDR: "wss://s9hlnv7y91.execute-api.eu-west-2.amazonaws.com"
       DEVICEINTELLIGENCETOGGLE: "0"
       BRANDREFRESHENABLED: "0"
+      FINGERPRINTCOOKIEDOMAIN: "account.gov.uk"
```
danacotoran, line 204: Is it okay if it's the same on pre-prod envs or should it be more specific?

danacotoran: Ah nevermind, just realised that was literally the whole point of this PR

alex9smith: Haha yes it is!

danacotoran (approved):

TaherKapasi (approved):

PR #2199, opened 2025-05-20
OLH-1704: Replace `csurf` with `csrf-sync`

alex9smith:
### What changed

`csrf-sync` uses the [Synchroniser Token Pattern](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#synchronizer-token-pattern) to do protection. It's mostly a drop in replacement for us.

It works by storing a token in the user's session. That's made available to the templates through `res.locals`. Any template that has a form has a hidden `_csrf` field which gets populated with that token. When the form is submitted the middleware checks for the token and validates that it matches the one in the session.

All that is exactly the same behaviour as before. The only difference is where the token is stored - in the session not in a cookie.

This is using one token per session, not one per request as can be configured. I tested using one per request but I was able to pretty easily get it to fail with aborting requests and then refreshing, It seems that I'd got through enough processing to generate a new token, but not enough for it to get saved into the DynamoDB session store by the time I ended the request. This will probably also be an issue with double submits or very slow page loads.

We had one token per session in the old implementation, so I don't see any issues with the same config here.

### Why did it change

This replaces our CSRF protection library as `csurf` has been deprecated.

### Related links

See other team's migrations - ([IPV](https://github.com/govuk-one-login/ipv-core-front/pull/1593) Core, [Auth](https://github.com/govuk-one-login/authentication-frontend/pull/2793))

### Sign-offs

- [x] New Node.JS/NPM dependencies have been signed-off by a Lead dev or Lead SRE 

## How to review
Run the frontend locally, go to a page with form, mess with the hidden CSRF field and submit the form. If it's working, you'll get redirected back to the security page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "body-parser": "^2.2.0",
         "connect-dynamodb": "^3.0.5",
         "cookie-parser": "^1.4.7",
-        "csurf": "^1.11.0",
+        "csrf-sync": "^4.2.1",
         "di-account-management-rp-registry": "github:govuk-one-login/di-account-management-rp-registry#main",
         "dompurify": "^3.2.5",
         "express": "^5.1.0",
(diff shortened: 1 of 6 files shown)


liegeandlief (approved):
That was a nice straightforward swap 🙌

alex9smith:
Yes, much easier than I was expecting in the end!

PR #2209, opened 2025-05-22
OLH-2328: Implement a early return if the user returns no services

latifcabinetoffice:
## Proposed changes

[OLH-2328] Implement a early return if the user returns no services

### What changed

Early return if the user has no services

### Why did it change

We have a number of error logs in production due to attempting to read a value that doesn't exist, ie "items".

"Your Services: failed with Error: No value defined: {}"

### Testing

Tested Locally to replicate the error log and the validated it not show up again post change.

### DIFF
--- a/src/utils/yourServices.ts
+++ b/src/utils/yourServices.ts
@@ -37,6 +37,10 @@ export const getServices = async (
       serviceStoreDynamoDBRequest(subjectId)
     );
 
+    if (!response["Item"]) {
+      return [];
+    }
+
     const services = unmarshallDynamoData(response["Item"])
       .services as Service[];
 


alex9smith:
That's a nice solution to clean up. I'm not sure how Sonar has given you a ✅ for 33% code coverage, but I will be the bad 👮🏻  . Could you add a unit test for this please?

latifcabinetoffice:
> That's a nice solution to clean up. I'm not sure how Sonar has given you a ✅ for 33% code coverage, but I will be the bad 👮🏻 . Could you add a unit test for this please?

thought I was blessed with existing coverage when it passed

saralk (approved):

PR #2223, opened 2025-05-28
OLH-2328: Keep original from url and log invalid from URL

latifcabinetoffice:
## Proposed changes

[OLH-2328] Adjust fromURL logic

### What changed

- Adjusted from URL logic to keep the original fromURL to prevent infinite depth fromURLs
- Changed the logger.warn for invalid fromURLs to include the invalid URL

### Why did it change

- Reduce bot traffic which is spamming the contact us page due to different URLs caused by fromURLs
- Better understand why we have a significant number of invalid URL errors in production

### How to review
Are there any issues with logging the from URL?

[OLH-2328]: https://govukverify.atlassian.net/browse/OLH-2328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/middleware/outbound-contact-us-links-middleware.ts
+++ b/src/middleware/outbound-contact-us-links-middleware.ts
@@ -2,7 +2,12 @@ import { NextFunction, Request, Response } from "express";
 import { PATH_DATA } from "../app.constants";
 
 export function buildUrlFromRequest(req: Request): string {
-  return `${req.protocol}://${req.get("host")}${req.originalUrl}`;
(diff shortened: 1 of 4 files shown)


alex9smith:
> Are there any issues with logging the from URL?

The full URL and any URL params already get logged in the 'page load' log events so I don't have any concerns with logging it again. There should never be any sensitive information in the URLs.

alex9smith (approved):

PR #2235, opened 2025-06-05
OLH-2721: Check if language selected is supported or fallback to en

latifcabinetoffice:
## Proposed changes

[OLH-2721] Check if language selected is supported or fallback to en

### What changed

Check language selection matches our supported options (en, cy) if not fallback to cy.

### Why did it change

Some users are passing a lng value of "" which is interpreted as an unknown language. As this value is not nullish it is not picked up by ?? and is passed to safeTranslate.  We could have changed it to || but with this change we now check against our list of supported languages.

### Related links

https://govukverify.atlassian.net/browse/OLH-2721

[OLH-2721]: https://govukverify.atlassian.net/browse/OLH-2721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/@types/express/index.d.ts
+++ b/@types/express/index.d.ts
@@ -30,7 +30,7 @@ declare module "express-serve-static-core" {
     i18n?: {
       language?: string;
     };
-    language?: string;
+    language?: LOCALE;
     t?: (string) => string;
     csrfToken?: () => string;
     oidc?: Client;
--- a/src/utils/safeTranslate.ts
+++ b/src/utils/safeTranslate.ts
@@ -11,7 +11,7 @@ export type translateResult =
 export const safeTranslate = (
   translate: TFunction<"translation", undefined>,
   key: string,
-  requestedLanguage: string,
+  requestedLanguage: LOCALE,
   options?: Record<string, any>
 ): translateResult => {
   const result: translateResult = translate(key, options);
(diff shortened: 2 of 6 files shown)


@types/express/index.d.ts
```diff
@@ -30,7 +30,7 @@ declare module "express-serve-static-core" {
     i18n?: {
       language?: string;
     };
-    language?: string;
+    language?: LOCALE.EN | LOCALE.CY;
```
liegeandlief, line 33: This can just be `language?: LOCALE` to save having to list out all of the languages

src/utils/safeTranslate.ts
```diff
@@ -11,7 +11,7 @@ export type translateResult =
 export const safeTranslate = (
   translate: TFunction<"translation", undefined>,
   key: string,
-  requestedLanguage: string,
+  requestedLanguage: LOCALE.EN | LOCALE.CY,
```
liegeandlief, line 14: Same here this can just be `requestedLanguage: LOCALE`

liegeandlief (commented):
I think you might be able to do away with all these changes by just updating the fallback language from `""` in `src/config/i18next.ts`

PR #2243, opened 2025-06-09
OLH-2691: Use an enum for audit event names

alex9smith:
### What changed

Switching the event names over to an enum means the type system will force us to only provide event names from the enum values, making it less likely we have a mistake eg. from a typo, and (I think) makes the intent clearer.

### Why did it change

We only want to allow the app to send audit events with names on a pre-defined list. Previously we had an object that mapped an allowed key to a string, but the functions that build and send audit events accepted any string as the event name. This reduces the risk of a mistake and forces any updates to the event service code to use the same pattern.


src/app.constants.ts
```diff
@@ -305,11 +305,10 @@ export const ENVIRONMENT_NAME = {
...
 
-export const EVENT_NAME = {
-  HOME_TRIAGE_PAGE_VISIT: "HOME_TRIAGE_PAGE_VISIT",
-  HOME_TRIAGE_PAGE_EMAIL: "HOME_TRIAGE_PAGE_EMAIL",
-};
-
+export enum EventName {
+  HOME_TRIAGE_PAGE_VISIT = "HOME_TRIAGE_PAGE_VISIT",
+  HOME_TRIAGE_PAGE_EMAIL = "HOME_TRIAGE_PAGE_EMAIL",
+}
```
liegeandlief, line 311: A lot of people don't recommend using enums in TypeScript for various reasons and using `as const` instead e.g.

```
export const EventName = {
  HOME_TRIAGE_PAGE_VISIT: "HOME_TRIAGE_PAGE_VISIT",
  HOME_TRIAGE_PAGE_EMAIL: "HOME_TRIAGE_PAGE_EMAIL",
} as const;
```

This is what I tend to do, but have no particularly strong feelings on it.

liegeandlief (approved):

src/app.constants.ts
alex9smith: Hmm ok. I wasn't aware that it was a controversial feature. Let me do some reading to see if any of those reasons would affect what I want to do here.

PR #2259, opened 2025-06-19
OLH-2796: add button role to links styled as buttons

liegeandlief:
### What changed

Use `govukButton` macro for link buttons to ensure they have the role `button`.

### Why did it change

To improve accessibility. When a link is styled as a button the user is likely to expect to be able to "click" the link using both the enter and space keys as is the case with HTML button elements. Without the `button` role links can only be "clicked" with the enter key. Adding the `button` role makes the link clickable with the space key too.

### Related links

https://govukverify.atlassian.net/browse/OLH-2796

### DIFF
--- a/src/components/common/confirmation-page/confirmation.njk
+++ b/src/components/common/confirmation-page/confirmation.njk
@@ -7,5 +7,5 @@
         titleText: heading
     }) }}
     <p class='govuk-body'>{{ message | safe }}</p>
-    <p><a class='govuk-button' href='{{ backLink }}'>{{ backLinkText }}</a></p>
+    <p><a class='govuk-button' role='button' href='{{ backLink }}'>{{ backLinkText }}</a></p>
 {% endblock %}
--- a/src/components/common/confirmation-page/confirmation.njk
+++ b/src/components/common/confirmation-page/confirmation.njk
@@ -8,12 +8,12 @@
         titleText: heading
     }) }}
     <p class='govuk-body'>{{ message | safe }}</p>
-    <p>{{ govukButton({
+    {{ govukButton({
         text: backLinkText,
         href: backLink,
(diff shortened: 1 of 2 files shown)


src/components/common/confirmation-page/confirmation.njk
```diff
@@ -1,11 +1,19 @@
...
+{% from "govuk/components/button/macro.njk" import govukButton %}
 {% block backLinkBlock %}
 {% endblock %}
 {% block pageContent %}
     {{ govukPanel({
         titleText: heading
     }) }}
     <p class='govuk-body'>{{ message | safe }}</p>
-    <p><a class='govuk-button' href='{{ backLink }}'>{{ backLinkText }}</a></p>
+    <p>{{ govukButton({
```
danacotoran, line 11: This is entirely unrelated to your changes (which seem good to me so I'm happy to approve this PR and investigate separately), but:
1. I'm pretty sure the `<p>` is unneeded here
2. Not sure if `{% block backLinkBlock %} {% endblock %}` is doing anything
3. More generally, I'm not sure why this additional confirmation page was added seeing as [another confirmation page/template already exists](https://github.com/govuk-one-login/di-account-management-frontend/blob/main/src/components/update-confirmation/index.njk) and seems like it should suit the intended purpose

danacotoran (approved):

src/components/common/confirmation-page/confirmation.njk
liegeandlief: I've removed the unneeded `p` tag.

The empty `backLinkBlock` is used to hide the back link at the top of the page:

Screenshot 2025-06-19 at 18 42 32

I'm not sure about the duplicate templates but I'll leave that for a separate investigation

PR #2268, opened 2025-06-23
OLH-2809: prevent res redirect on intent from user passed variable

latifcabinetoffice:
### What changed

- Check intent before redirecting user
- Throw a badRequestError if a unknown intent is passed

### Why did it change

Prevent user controlled actions controlling the intent

### DIFF
--- a/src/components/resend-phone-code/tests/resend-phone-code-controller.test.ts
+++ b/src/components/resend-phone-code/tests/resend-phone-code-controller.test.ts
@@ -97,6 +97,8 @@ describe("resend phone code controller", () => {
 
       try {
         await resendPhoneCodePost(fakeService)(req as Request, res as Response);
+        expect(fakeService.sendPhoneVerificationNotification).not.to.have.been
+          .called;
       } catch (err) {
         expect(err).to.be.instanceOf(BadRequestError);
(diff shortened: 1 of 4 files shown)


src/components/resend-phone-code/tests/resend-phone-code-controller.test.ts
```diff
@@ -79,5 +80,26 @@ describe("resend phone code controller", () => {
...
+
+      res.locals.sessionId = "123456-djjad";
+      req.session.user.tokens = { accessToken: "token" } as any;
+      req.body.phoneNumber = "+33645453322";
+      req.body.intent = "unknown";
+      req.session.user.email = "test@test.com";
+      req.session.user.state = { changePhoneNumber: getInitialState() };
+
+      try {
+        await resendPhoneCodePost(fakeService)(req as Request, res as Response);
```
liegeandlief, line 99: Probably worth putting an assertion after the function call too, otherwise this test will pass even when the function doesn't throw.

e.g. `expect(fakeService.sendPhoneVerificationNotification).not.to.have.been.called;`

liegeandlief (changes requested):

PR #2273, opened 2025-06-24
OLH-2783: Use a session scoped session cookie

alex9smith:
### What changed

Remove the max age definition from our session cookie config. This means express-session won't set an `Expires` field on the cookie and therefore browsers will treat our session cookie as scoped to that browser session.

In practice, this means that we'll log users out when they close the last tab / window with Home open in it.

- [x] Added to local startup config

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -5,7 +5,6 @@ OIDC_CLIENT_ID=
 OIDC_CLIENT_SCOPES=openid phone email am offline_access govuk-account
 AM_API_BASE_URL=https://am-stub.home.dev.account.gov.uk
 AM_YOUR_ACCOUNT_URL=http://localhost:6001
-SESSION_EXPIRY=1800000
 SESSION_SECRET=secret
 GOV_ACCOUNTS_PUBLISHING_API_URL=http://localhost:4444
 GOV_ACCOUNTS_PUBLISHING_API_TOKEN=token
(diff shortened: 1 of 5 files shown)


liegeandlief (commented):
If the user doesn't close their browser then their session will last as long it exists in DynamoDB. Where is the expiry of sessions in DynamoDB configured? Do we need to check and maybe adjust that expiry time

alex9smith:
Hmm good point. We do have the TTL configured on that table but I'm not sure how this `expires` object is set. 

https://github.com/govuk-one-login/di-account-management-frontend/blob/6963e6a59e7f624ef55b242eae263e9ecb6d47bc/deploy/template.yaml#L1960-L1962

alex9smith:
It's set by the library we're using and will default to "now plus a day" if our session cookie doesn't have an expiry time. That might not work for us 😞 

https://github.com/ca98am79/connect-dynamodb/blob/eeeb8d384fd71e7571f6c586bee0813b3e9d7702/lib/connect-dynamodb.js#L340-L352

liegeandlief:
> It's set by the library we're using and will default to "now plus a day" if our session cookie doesn't have an expiry time. That might not work for us 😞
> 
> https://github.com/ca98am79/connect-dynamodb/blob/eeeb8d384fd71e7571f6c586bee0813b3e9d7702/lib/connect-dynamodb.js#L340-L352

Frustrating that the cookie and session expiry can't be configured independently 😖

alex9smith:
I've [opened a PR](https://github.com/ca98am79/connect-dynamodb/pull/103) with the feature - we shall see!

PR #2278, opened 2025-06-25
OLH-2813: Confirmation page duplication

danacotoran:
### What changed
There are currently two update confirmation templates serving the same purpose.
This replaces references to the "new" one added for MFA MM with the old/preexisting one.
As far as I could tell the outcome is identical but do correct me if I missed anything.

| Before  | After |
| backup-text-message-confirmation-before  | backup-text-message-confirmation-after  |
| backup-auth-app-confirmation-before  | backup-auth-app-confirmation-after  |
| remove-backup-confirmation-before  | remove-backup-confirmation-after  |
| switch-methods-confirmation-before  | switch-methods-confirmation-after  |
| change-default-method-confirmation-before  |  change-default-method-confirmation-after  |

### Why did it change
Avoid repeating ourselves.

## How to review
Check these templates and verify that I didn't unintentionally break something.


liegeandlief (approved):
Looks good, a couple of small comments but nothing blocking approval

PR #2284, opened 2025-06-26
Olh 2506: searchable list updates

danacotoran:
### What changed

Add "Clear search" link to "Services you can use with GOV.UK One Login" page.

Update page title to reflect current search term.

Add missing Welsh translations for searchable list

Change input field type from `search` to `text` as a quick fix for the inline "x" which is appended to the field in some browsers – this can also be done with CSS but I figured changing the type of field would be easier and would avoid increasing CSS payload, with minimal consequence:
- when AT is used, the form still appears in the landmarks list as a `search` landmark and functions the same way
- there is a subtle difference in the way the field is announced when using screen readers ("text" vs "search"). However the purpose of the field is clear from the label/context
- there are slight differences in mobile device keyboard layout presentation – some examples below (in hindsight maybe [enterkeyhint](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/enterkeyhint) could be used here, for a more consistent experience) 

| type=search  | type=text |
| typesearch  | typetext  |
| typesearch3  | typetext3  |

### Why did it change
Bring searchable list page in line with the specification in Figma (linked on the original ticket)

The original task centred around "Clear search", however the scope of the ticket increased slightly as I kept spotting bits to fix. 

### Sign-offs

## How to review
Content/design has been signed off.
Check everything looks as described on `/services-using-one-login`. Feel free to share any concerns or issues with the approach.

### DIFF
--- a/src/components/search-services/index.njk
+++ b/src/components/search-services/index.njk
@@ -45,7 +45,7 @@
           classes: "govuk-!-margin-bottom-3"
         },
         id: "search_box",
-        type: "search",
+        type: "text",
         name: "q",
         classes: "search-services__input",
         value: query
--- a/src/components/search-services/index.njk
+++ b/src/components/search-services/index.njk
@@ -2,7 +2,11 @@
 {% set hideAccountNavigation = true %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/input/macro.njk" import govukInput %}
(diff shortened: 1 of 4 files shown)


liegeandlief (approved):

danacotoran:
Ah, good spot @liegeandlief! I'll bring that up to Mel and address it as a separate bit of work if she reckons it's an issue.

PR #2303, opened 2025-07-02
OLH-2768: remove clearCookies function

liegeandlief:
### What changed

Removes the `clearCookies` function.

### Why did it change

It was suspected that the function wasn't working and cookies weren't being removed. On further investigation it turns out that the function was working but it is not necessary.

The function was only used to remove the `am` cookie. We don't need to manually remove the `am` cookie. When the express session is destroyed the user will get a new session on their next request and their existing `am` cookie (which no longer corresponds to a session) is replaced with a new `am` cookie containing the session ID for their new session.

https://github.com/user-attachments/assets/126342bc-a430-433c-9ffb-a4e519e7d8ab

## How to review

- Go to the dev deployment and check that you can log in and carry out actions
- Make a note of your `am` cookie value.
- Logout
- On the logout confirmation page check that your `am` cookie value has changed

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -304,9 +304,9 @@ Mappings:
       SUPPORTSEARCHABLELIST: "0"
       SHOWCONTACTEMERGENCYMESSAGE: "0"
       METHODMANAGEMENTBASEURL: "https://z7lornzyy5-vpce-0e594accb3d775457.execute-api.eu-west-2.amazonaws.com/integration/v1"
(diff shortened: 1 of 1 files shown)


TaherKapasi (approved):
Hi @liegeandlief - Would it possible to capture some screenshot evidence that the am cookie gets replaced when revisting and is cleared when session is destroyed

liegeandlief:
@TaherKapasi I've added a screen recording to the description

PR #2331, opened 2025-07-10
OLH-2898: refresh token improvements

liegeandlief:
### What changed

Removes the refresh token middleware.

Adds a refresh token function which is called when the access token is used (i.e. when making API calls).

This means we don't have to remember to add a refresh token middleware when a route makes API calls. The token is refreshed when it is used and the code making an API call doesn't have to concern itself with it.

### Why did it change

To stop us having to remember to add the refresh token middleware when a route makes API calls and to abstract the token refresh logic away from the API caller.


src/utils/oidc.ts
```diff
@@ -90,10 +93,55 @@ const clientAssertionGenerator = (
...
+        );
+        req.session.user.tokens.accessToken = tokenSet.access_token;
+        req.session.user.tokens.refreshToken = tokenSet.refresh_token;
+      } catch (error) {
+        req.metrics?.addMetric(
+          "refreshTokenMiddlewareError",
+          MetricUnit.Count,
+          1
+        );
+        throw error;
```
latifcabinetoffice, line 131: Should we redirect the user to sign in again?

latifcabinetoffice (changes requested):

PR #2344, opened 2025-07-14
OLH-2897: Clear users' state on the add MFA confirmation pages

alex9smith:
### What changed

Clear users' state on the add MFA confirmation pages.

### Why did it change
We were seeing requests to the API that were missing the new phone number when adding a new SMS method, and errors that users trying to add a new app method already had two methods.

Both of these are because users were getting to the confirmation page (and successfully adding the new backup method) and then using the browser's back button to get back to the security page. This repeated the POST to create the new method, which caused the frontend to make a bad API call.

The state machine should have prevented these requests from being made and redirected the user back to the security page instead, but we'd missed clearing the state on the confirmation page. That meant the state machine still viewed the user as being 'in the add backup journey' and allowed the repeat request through.

### Related links

This addresses both [OLH-2897](https://govukverify.atlassian.net/browse/OLH-2897) and [OLH-2896](https://govukverify.atlassian.net/browse/OLH-2896). These are the same root cause but showed up as different errors on the SMS and app journeys.


src/components/check-your-phone/check-your-phone-controller.ts
liegeandlief, line 1: If think there is also another log to remove from this file in the POST handler

liegeandlief (changes requested):

src/components/check-your-phone/check-your-phone-controller.ts
alex9smith: 😴 Thanks

PR #2359, opened 2025-07-17
OLH-2769: remove MFA feature flags

liegeandlief:
## What changed

Removes all of the MFA feature flags and code made redundant by their removal.

## Why did it change

MFA is here to stay!

## How to review

Please check out the branch and do some MFA sanity testing to check I haven't broken anything


src/components/security/index.njk
```diff
@@ -51,86 +51,54 @@
...
-                {% endfor %}
-                {% if not foundSecondary %}
-                  <h3 class='govuk-heading-s'>{{ 'pages.security.mfaSection.supportChangeMfa.backup.headingNoBackup' | translate }}</h3>
-                  {# 2: backup method does not exist => display option to add a new backup method #}
-                  <p class="govuk-body">{{ 'pages.security.mfaSection.supportChangeMfa.backup.paragraph' | translate }}</p>
-                  <p class="govuk-body">{{ 'pages.security.mfaSection.supportChangeMfa.backup.paragraph2' | translate }}</p>
-                  <p class="govuk-body">
-                    <a href="{{enterPasswordUrl}}&type=addBackup" class="govuk-link">{{ 'pages.security.mfaSection.supportChangeMfa.backup.link' | translate }}</a>
-                  </p>
+      {# if changing authenticator app is supported, show the new design #}
```
alex9smith, line 54: Can get rid of this comment too

.github/workflows/build-test-application.yml
```diff
@@ -88,3 +88,4 @@ jobs:
       SESSION_SECRET: secret
       SESSION_EXPIRY: 3600000
       SESSION_STORE_TABLE_NAME: account-mgmt-frontend-SessionStore
+      METHOD_MANAGEMENT_BASE_URL: https://method-management-v1-stub.home.dev.account.gov.uk/v1
```
alex9smith, line 91: Does this actually get called at any point during the tests? If so, we should use build not dev as it's much less likely to be broken

src/components/check-your-phone/check-your-phone-controller.ts
```diff
@@ -145,134 +128,119 @@ export function checkYourPhoneGet(req: Request, res: Response): void {
...
+  const changePhoneNumberResult = await changePhoneNumberwithMfaApi(
+    mfaClient,
+    code,
+    intent,
+    newPhoneNumber,
+    res.locals.trace,
+    req.session.mfaMethods,
+    req.t
+  );
+  isPhoneNumberUpdated = changePhoneNumberResult.success;
```
alex9smith, line 220: I think we can get rid of this `isPhoneNumberUpdated` variable completely now and move the success state up to the top to get the early return in first.

alex9smith (changes requested):
What a satisfying PR! Just a couple of small requests.

PR #2382, opened 2025-07-29
OLH-2959: Add TEST phase banner in pre-prod

danacotoran:
### What changed

Add TEST version of phase banner to mirror auth team's implementation for their pre-prod journeys.

The auth journeys have an additional yellow bar under the header however this didn't look great with the secondary navigation in home, and I thought that the phase banner update made the current environment clear enough. 

~Added a layout test which checks for the existence of basic page furniture elements including the phase banner.~ I've removed the bonus test after spending hours trying to debug why it was failing 🫠  It might be more practical to use @liegeandlief's Playwright implementation to verify the existence of page furniture elements (including the phase banner) instead. 

### Why did it change
Consistency with the auth journeys. Make it immediately clear the current environment is non-prod.

## How to review
When this branch is tested locally (or in any env other than prod), the phase banner tag should be yellow, and the text should reflect that the current env is a test one.

Like the screenshot below, but minus the yellow bar. 

Screenshot 2025-07-24 at 16 02 55

### DIFF
--- a/src/assets/scss/application.scss
+++ b/src/assets/scss/application.scss
@@ -387,13 +387,16 @@ $nav-button-padding-bottom: $nav-link-padding + $nav-link-outline-thickness;
   // https://accessibility.blog.gov.uk/2017/03/27/how-users-change-colours-on-websites/
   outline: 2px solid transparent;
   outline-offset: -2px;
-
-  // the most important properties for keeping the phase banner looking the same have been !important-ed
-  color: govuk-colour("white") !important;
-  background-color: govuk-colour("blue") !important;
   letter-spacing: 1px !important;
   text-decoration: none !important;
   text-transform: uppercase !important;
+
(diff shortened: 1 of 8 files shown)


src/assets/scss/application.scss
```diff
@@ -387,13 +387,16 @@ $nav-button-padding-bottom: $nav-link-padding + $nav-link-outline-thickness;
...
   outline-offset: -2px;
-
-  // the most important properties for keeping the phase banner looking the same have been !important-ed
-  color: govuk-colour("white") !important;
-  background-color: govuk-colour("blue") !important;
   letter-spacing: 1px !important;
   text-decoration: none !important;
   text-transform: uppercase !important;
+
+  // the most important properties for keeping the phase banner looking the same have been !important-ed
```
liegeandlief, line 394: I like your verbing of `!important-ed`

liegeandlief (approved):

PR #2415, opened 2025-08-06
OLH-2955: fix session redirect race condition

liegeandlief:
Monkey-patch res.redirect to defer calling it until after the session has been saved. This prevents race conditions where the user follows a redirect response and the subsequent request reads the session before the changes to the session made in the previous request have been saved.

The fix has been tested by using our integration testing setup to make many simultaneous requests to the change email address journey which in several places writes to the session before redirecting. Before this change many of these requests failed due to race conditions. With this change all the requests succeed.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -96,6 +96,7 @@ const APP_VIEWS = [
 
 async function createApp(): Promise<express.Application> {
   const app: express.Application = express();
+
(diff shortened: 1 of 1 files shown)


alex9smith (approved):
Great, something like this was exactly what I was hoping we'd be able to do

PR #2431, opened 2025-08-13
BAU: add skipTarget-local for webchat test

danacotoran:
The webchat test is failing for some of us locally although it is passing in CI. This appears to be related to the websocket connection errors we have been observing for a while.

See webchat test timing out locally due to websocket CSP issue. The same error doesn't happen in higher environments. 
Screenshot 2025-08-13 at 13 15 45

This disables the webchat test locally (hopefully as a temporary measure). Tests will still be running in the GH action. 
Also added `@postDeploy` and removed the cookie step as per Michael's suggestion.

The cookie acceptance step is unnecessary as the contact page functions the same whether cookies have been accepted or not, and there is a separate cookie consent banner test.


integration-tests/tests/features/@contact-govuk-one-login.feature
```diff
@@ -1,10 +1,10 @@
+@postDeploy
 Feature: "Contact GOV.UK One Login" page
 
   @failMobile
   Scenario: Visit the "Contact GOV.UK One Login" page
     Given I go to the "Contact GOV.UK One Login" page
     And the page has finished loading
-    And I accept cookies
```
liegeandlief, line 7: I think maybe we should leave the step here just so the banner isn't in the screenshot of this scenario

integration-tests/tests/features/@contact-govuk-one-login.feature
```diff
@@ -1,10 +1,10 @@
+@postDeploy
```
liegeandlief, line 1: I think this could be put on the two individual webchat scenarios rather than the whole feature. My intention with the integration testing changes was to shift the testing left as much as possible. The first scenario is passing fine locally and in CI and I don't think there's a need to run it post-deploy too?

liegeandlief (approved):

PR #2441, opened 2025-08-20
OLH-2969: fix code smells

liegeandlief:
Fixes code smells reported by SonarCloud

### DIFF
--- a/src/components/report-suspicious-activity/report-suspicious-activity-controller.ts
+++ b/src/components/report-suspicious-activity/report-suspicious-activity-controller.ts
@@ -82,9 +82,13 @@ export async function reportSuspiciousActivityGet(
     res.render("common/errors/404.njk");
   } else {
     try {
-      assert.ok(
-        req.query.event !== undefined && req.query.event !== "",
+      assert(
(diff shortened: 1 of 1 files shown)


src/components/report-suspicious-activity/tests/report-suspicious-activity-controller.test.ts
```diff
@@ -346,12 +365,32 @@ describe("report suspicious activity controller", () => {
...
+        res as Response,
+        () => {}
       );
 
       // Assert
       expect(res.status).to.have.been.calledOnceWith(404);
       expect(res.render).to.have.been.calledOnceWith("common/errors/404.njk");
     });
+
+    it("Should call next when report parameters don't pass validation", async () => {
```
TaherKapasi, line 377: Hi @liegeandlief - this test name is exactly the same as the own above - but it has a different req.query attribute, could we name them instead

"Should call next when the **event** query string parameter fails validation"
"Should call next when the **page** query string parameter fails validation"

TaherKapasi (changes requested):

TaherKapasi (approved):

PR #2443, opened 2025-08-20
BAU: template refactoring

danacotoran:
### What changed
I changed the spacing/indentation on a few templates which makes the changes in this PR look more extensive than they actually are.

There are 3 main categories of changes:

### Changed all templates over to `base-page.njk` 
The `base-page.njk` template extends the main `base.njk` template and ensures a two-thirds layout is applied across all pages. A number of templates were extending  `base.njk` and applying the two-third layout manually using `govuk-grid-column-two-thirds`. This reduces some of that repetition and ensures all pages are using the correct layout. 
### Updated `govuk-heading-l`
I noticed instances of `.govuk-heading-l` usually had the `.govuk-!-margin-top-0` override class. But this was redundant, as this style of heading [already has margin-top:0 as defined in govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/88f8b7a1780caf63e0d9fb43c46cefeba928017e/packages/govuk-frontend/src/govuk/core/_typography.scss#L24).

One remaining inconsistency in the header styles is the `govuk-!-margin-bottom-3`. It can be argued that this is also unneeded, and the default DS styles for govuk headers can be used.  However this should be raised with the UCD team first, so I decided against including it as part of this PR.  

### Refactored buttons 
Remove instances of type:"Submit" as this is redundant (the govukButton component has `type="submit"` by default unless otherwise specified).
Remove instances of `button_text` and `default` filter as:
1. we don't declare a `button_text` anywhere so the text of these buttons always defaults to the translation string anyway
2. all the specified translation keys for buttons do exist in the translation files, and there isn't currently a need to support
overriding them with the way our templating is currently set up.

### Why did it change

Although there is still room for improvement, this reduces some repetition and retires bits of code which are no longer needed.

## How to review
The commits are granular-ish so it might be best to sense check this on a commit-by-commit basis.

### DIFF
--- a/src/components/activity-history/index.njk
+++ b/src/components/activity-history/index.njk
@@ -23,7 +23,7 @@
 {% endfor %}
 
 {% block pageContent %}
-  <h1 class="govuk-heading-l govuk-!-margin-top-0">{{ 'pages.activityHistory.header' | translate }}</h1>
+  <h1 class="govuk-heading-l">{{ 'pages.activityHistory.header' | translate }}</h1>
   {% if data.length %} 
     <p class="govuk-body">{{ 'pages.activityHistory.explanation' | translate }}</p>
     {% if hasEnglishOnlyServices and currentLngWelsh %}<p class="govuk-body" data-test-id="no-welsh-notice">{{ 'pages.activityHistory.noWelshNotice' | translate }}</p>{% endif %}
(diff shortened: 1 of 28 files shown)


danacotoran:
Thanks for approving @liegeandlief, I haven't merged it in yet as it's a Friday afternoon and I'm kind of getting cold feet in case something goes wrong (although none of the changes should be breaking, really).

I will solve the merge conflict on it today but wait until next week to merge it in perhaps.

PR #2457, opened 2025-08-28
OLH-3057: Make users enter their password before global logout 

alex9smith:
### What changed

Make users enter their password before global logout.

This journey is slightly different to the others in Home as we want to show the global logout page first, then have the user enter their password.

That means the series of HTTP requests that make up this journey now look like:

```
GET  /security - click to "sign out of all apps"
GET  /global-logout - click the red button
POST /global-logout - redirect to password page
GET  /enter-password - user enters password
POST /enter-password - validates password, redirect to /global-logout/confirm
GET  /global-logout/confirm - sends the audit event, logs the user out
of Home
GET  / - redirects to sign in
... - user sees the sign in page

```

I chose to add a new route that contains the logout behaviour. That has a couple of benefits:

1. It keeps the enter password controller behaviour the same as all other journeys
2. We can use the same state checks on the new route to make sure users have re-authenticated before they logout

The downside is that we have a destructive action triggered from a GET request which isn't in the spirit of semantic HTTP. I think the benefits outweigh this slight concern though. I've used (and tested) the state machine to prevent users from calling this endpoint without re-authenticating first.

> [!NOTE]
> The content on the enter password screen is not final.
> We'll need to update this and get a Welsh translation before going live.

### Why did it change

We want to force the user to re-authenticate as this action will delete credentials saved in the GOV.UK App.

### Testing

- [x] Automated test coverage includes features that are flagged off


src/locales/en/translation.json
```diff
@@ -368,6 +368,10 @@
         "header": "Enter your password",
         "paragraph": "We need to make sure it’s you before you can manage how you get security codes."
       },
+      "globalLogout": {
+        "header": "Enter your password",
+        "paragraph": "We need to make sure it’s you before you can manage how you get security codes."
```
TaherKapasi, line 373: Security codes? Does this content need to be updated something like "We need to make sure it’s you before you can sign out of all devices?

src/locales/cy/translation.json
```diff
@@ -291,6 +291,10 @@
         "header": "Rhowch eich cyfrinair",
         "paragraph": "Mae angen i ni wneud yn siŵr mai chi yw chi cyn i chi allu rheoli sut rydych yn cael codau diogelwch."
       },
+      "globalLogout": {
+        "header": "Rhowch eich cyfrinair",
+        "paragraph": "Mae angen i ni wneud yn siŵr mai chi yw chi cyn i chi allu rheoli sut rydych yn cael codau diogelwch."
```
TaherKapasi, line 296: Security codes? Does this content need to be updated something like "Mae angen i ni sicrhau mai chi ydyw cyn y gallwch chi allgofnodi o bob dyfais.?

integration-tests/snapshots/Desktop/tests/features/@global-logout.feature.spec.js/Global-logout-Trigger-a-global-logout-1.png
TaherKapasi, line 1: Is it possible to get the password challenge page with the text And is it possible to get both English and Welsh screenshots?

TaherKapasi (changes requested):

integration-tests/tests/features/@global-logout.feature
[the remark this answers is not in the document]
alex9smith: >Temporary Suspended (AIS)-> allowed to Globally logged out

Is this the right behaviour? For all other journeys, if a user has a temporary suspension we also log them out and show them the temporarily unavailable page. 

I'll add some tests for both suspension cases either way. The password is incorrect one is trickier as we don't have a scenario in the stub for a user with an incorrect password. I could look at that in a followup PR if you think it's worth it?

integration-tests/snapshots/Desktop/tests/features/@global-logout.feature.spec.js/Global-logout-Trigger-a-global-logout-1.png
alex9smith: I've added a screenshot check for the password screen. I'm not sure about Welsh but if we do do that we'll want to check English and Welsh for all the tests. That's not going to be simple as our assertions use the content of the page and page titles a lot, so we should tackle it in a separate ticket.

integration-tests/tests/steps/shared.ts
```diff
@@ -66,6 +71,16 @@ Then("the page title is {string}", async ({ page }, pageTitle: string) => {
   expect(await page.title()).toBe(pageTitle);
 });
 
+Then("I am on the sign in page", async ({ page }) => {
+  await expect(
+    page.getByText("API Simulation Tool", { exact: true })
+  ).toBeVisible();
+});
+
+Then("the page contains {string}", async ({ page }, content: string) => {
```
liegeandlief, line 80: the nittiest of nits but could this step be reworded to `the page contains the text {string}`?

src/components/enter-password/enter-password-controller.ts
```diff
@@ -43,6 +43,7 @@ const REDIRECT_PATHS: Record<UserJourney, string> = {
   [UserJourney.RemoveBackup]: PATH_DATA.DELETE_MFA_METHOD.url,
   [UserJourney.SwitchBackupMethod]: PATH_DATA.SWITCH_BACKUP_METHOD.url,
   [UserJourney.ChangeDefaultMethod]: PATH_DATA.CHANGE_DEFAULT_METHOD.url,
+  [UserJourney.GlobalLogout]: PATH_DATA.GLOBAL_LOGOUT_CONFIRM.url,
 };
 
 const getOplValues = (): OplSettingsLookupObject => ({
```
liegeandlief, line 49: Is there a separate Jira ticket for adding the content ID for this journey?

integration-tests/snapshots/Desktop/tests/features/@global-logout.feature.spec.js/Global-logout-Trigger-a-global-logout-1.png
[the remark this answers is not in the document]
alex9smith: Ah yes, I renamed the first scenario when I added more. Will remove.

integration-tests/tests/features/@global-logout.feature
```diff
@@ -0,0 +1,58 @@
+Feature: Global logout
+
+  @failMobile
+  Scenario: A user triggers a global logout
+    Given I go to the "Root" page
+    And I sign in as the "default" user
+    And the page has finished loading
```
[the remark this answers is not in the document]
alex9smith: Will do. From the context of the other places it's used (eg. in the [security feature](https://github.com/govuk-one-login/di-account-management-frontend/blob/main/integration-tests/tests/features/%40security.feature#L8)) I thought it might be needed to follow the redirect chain after logging in, but I've tested these without it and they all pass still.

src/components/enter-password/enter-password-controller.ts
alex9smith: I have a checklist item on [this ticket](https://govukverify.atlassian.net/browse/OLH-3057) to create that follow up task

liegeandlief (approved):

PR #2475, opened 2025-09-10
OLH-3135: reduce snapshot image quality

liegeandlief:
### What changed

Reduces the quality of screenshots

### Why did it change

To prevent the repo from getting too big


alex9smith:
@liegeandlief I think the test failures are because of the new JPEG artifacts. We might need to go up in quality a bit

PR #2486, opened 2025-10-01
OLH-767 add caching header and cachebusting for static assets

saralk:
### What changed

I changed the approach, as creating the hash at run time was causing timeouts.

- Generate a hash of the public folder at build time, store this result in a static-hash.json file in the dist folder.
- Use that hash in the URLs when requesting static assets
- Set a max age of one year on the static assets folder (when not running locally)

The `/assets` folder (e.g. the assets from the govuk frontend library) don't have versioning on the government crest image, but it does version the font files, which are our largest assets. So we think it's worth the trade off.

Currently, caching is disabled on cloudfront, so the caching will only happen further downstream. We still expect this to make a meaningful difference, as it will stop the assets being re-requested every time the user requests a new page.

### Why did it change

Currently our server handles all the requests, including static assets. This was problematic as every page request was requesting all the static assets, which was increasing load on our server.

### Testing

I have tested this by running the app, and verifying that the cachebusting property exists, and the headers are set correctly. Then I manually made a change to the static files, and saw that the hash changed.

### Sign-offs

### Monitoring


src/app.ts
```diff
@@ -121,7 +121,8 @@ async function createApp(): Promise<express.Application> {
   app.use(
     "/assets",
     express.static(
-      path.resolve("node_modules/govuk-frontend/dist/govuk/assets")
+      path.resolve("node_modules/govuk-frontend/dist/govuk/assets"),
+      { maxAge: "1y" }
```
liegeandlief, line 125: Did you mean to add this caching? It seems to contradict the PR description

src/config.ts
```diff
@@ -91,6 +91,10 @@ export function getAwsRegion(): string {
   return process.env.AWS_REGION ?? "eu-west-2";
 }
 
+export function getStaticAssetsPath(): string {
+  return process.env.STATIC_ASSETS_PATH || "../public/";
```
liegeandlief, line 95: What is the need for `STATIC_ASSETS_PATH`? Will it not always be `public` across all environments?

liegeandlief (changes requested):

liegeandlief (approved):

liegeandlief (approved):

PR #2892, opened 2026-04-14
OLH-3699 Create Passkey Flow

saralk:
### What changed

- Enable passkeys in the staging environment
- Create the start point for the passkey creation flow, handing over to account components
- Use the state machine on the entire flow (so that users need to enter a password)
- Add unit and integration tests for the above
- Some small improvements to the TypeScript typing.

### Testing

### Sign-offs

### Monitoring


src/utils/initiateAmcRedirect.ts
```diff
@@ -29,7 +29,7 @@ export async function initiateAmcRedirect(
     email,
   });
 
-  const redirectUri = `${getHomeBaseUrl()}${PATH_DATA.AMC_CALLBACK.url}`;
+  const redirectUri = `${getAmcBaseUrl()}${PATH_DATA.AMC_CALLBACK.url}`;
```
liegeandlief: This redirect URL is the callback URL which AMC redirects back to on completion so should be a Home URL not an AMC URL. I think the previous use of `getHomeBaseUrl` was correct

liegeandlief (changes requested):

src/utils/initiateAmcRedirect.ts
saralk: AMC_BASE_URL is a new environment variable, it lets us override the redirect url when we run the integration tests. Otherwise it will redirect to home.dev and the tests fail

https://github.com/govuk-one-login/di-account-management-frontend/pull/2892/changes#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R256

liegeandlief: Ah okay. I wonder if the environment variable and config function could be named more clearly. The current name makes it sound like it is an AMC URL. Maybe `AMC_CALLBACK_BASE_URL` and `getAmcCallbackBaseUrl`

saralk: yeah makes sense, udpated

src/components/amc-callback/amc-callback-utils.ts
```diff
@@ -166,12 +167,14 @@ export async function handleJourneyOutcomeResponse(
...
+  req.session.createdPasskeyAaguid = currentJourney?.details?.aaguid;
+
   if (success && isPasskeyJourney) {
-    return res.redirect("/todo-passkey-confirmation");
+    return res.redirect(PATH_DATA.PASSKEY_CREATED_CONFIRMATION.url);
   } else if (!success && error?.code === 1001) {
     await handleLogout(req, res, LogoutState.AmcSignedOut);
   } else if (!success && isPasskeyJourney && error?.code === 1002) {
-    return res.redirect("/todo-user-aborted");
+    return res.redirect(PATH_DATA.YOUR_SERVICES.url);
```
liegeandlief: This should go to the new sign in details page, not the your services page

src/components/create-new-passkey/create-new-passkey-controller.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export async function createNewPasskeyGet(
+  req: Request,
+  res: Response
+): Promise<void> {
+  const mfaClient = await createMfaClient(req, res);
+  const passkeys = await mfaClient.getPasskeys();
+
+  if (passkeys.data.length >= maxNumberOfPasskeys) {
+    res.redirect(PATH_DATA.START.url);
```
liegeandlief: This should also go to the new sign in details page

src/components/update-confirmation/update-confirmation-controller.ts
```diff
@@ -308,3 +310,38 @@ export async function changeDefaultMethodConfirmationGet(
...
+    ? req
+        .t("pages.createPasskeyConfirmation.summaryHtml1")
+        .replace("[passkeyName]", passkeyName)
+        .concat(baseHtml)
+    : baseHtml;
+
+  setOplSettings(
+    {
+      ...PASSKEY_COMMON_OPL_SETTINGS,
+      contentId: "d9c3e311-1c8b-4a5c-9cbd-7f0a1b2e3110",
```
liegeandlief: Where did this content ID come from? I don't think we have the content IDs for passkeys journeys yet

src/locales/cy/translation.json
liegeandlief, line 1: There is a separate ticket to add all of the Welsh translations once they're completely ready. To save confusion in future can these translations be removed for now and then we can do them all in one go when they all ready

src/utils/initiateAmcRedirect.test.ts
```diff
@@ -21,6 +21,7 @@ vi.mock("../config.js", () => ({
   getAmcClientId: vi.fn(() => "test-client-id"),
   getHomeBaseUrl: vi.fn(() => "https://home.example.com"),
   getServiceDomain: vi.fn(() => "example.com"),
+  getAmcBaseUrl: vi.fn(() => "https://home.example.com"),
```
liegeandlief: Does this need to be changed to `getAmcCallbackBaseUrl`?

src/utils/opl.ts
```diff
@@ -29,6 +29,9 @@ export const DELETE_ACCOUNT_COMMON_OPL_SETTINGS: Partial<OplSettings> = {
 export const ACTIVITY_COMMON_OPL_SETTINGS: Partial<OplSettings> = {
   taxonomyLevel2: "activity",
 };
+export const PASSKEY_COMMON_OPL_SETTINGS: Partial<OplSettings> = {
+  taxonomyLevel2: "passkey",
+};
```
liegeandlief: This hasn't been decided upon yet and there is a separate ticket for it. Please can it be removed

src/app.constants.ts
```diff
@@ -199,8 +204,17 @@ export const PATH_DATA: Record<
   LOGOUT_REDIRECT: {
     url: "/logout-redirect",
   },
-  AMC_CALLBACK: { url: "/amc/callback" },
+  AMC_CALLBACK: {
+    url: "/amc/callback",
+    event: EventType.ValueUpdated,
+    type: UserJourney.CreatePasskey,
```
liegeandlief, line 210: This works while we only have one AMC journey but what about when we have multiple?

liegeandlief (changes requested):
Just a few small issues but otherwise this looks great

src/app.constants.ts
saralk: Yeah, that's a good point, we might need to extend the state machine to handle an array of types. In my opinion it's out of scope of this ticket

liegeandlief (approved):

PR #3007, opened 2026-05-27
OLH-4127: send passkey delete audit events

martin-gwar:
## Proposed changes

[OLH-4127] send passkey delete audit events

Two new audit event names added to the EventName enum:

HOME_PASSKEY_DELETE_SUCCESSFUL

HOME_PASSKEY_DELETE_FAILED

### Testing

### Sign-offs

### Monitoring

## How to review

[OLH-4127]: https://govukverify.atlassian.net/browse/OLH-4127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/event-service.ts
```diff
@@ -205,6 +205,8 @@ export function eventService(
         break;
 
       case EventName.HOME_GLOBAL_LOGOUT_REQUESTED:
+      case EventName.HOME_PASSKEY_DELETE_SUCCESSFUL:
```
liegeandlief, line 210: Looking at the event specs I think you need to set the extension `journey-type` to `ACCOUNT_MANAGEMENT` for these two event types

liegeandlief (changes requested):
One small change and also looks like there are some Sonar code smells to fix

src/services/event-service.ts
martin-gwar: should it be for all three of them like this: 
      case EventName.HOME_GLOBAL_LOGOUT_REQUESTED:
      case EventName.HOME_PASSKEY_DELETE_SUCCESSFUL:
      case EventName.HOME_PASSKEY_DELETE_FAILED:
        baseEvent.extensions = {
          "journey-type": "ACCOUNT_MANAGEMENT",
        };
        break;

liegeandlief: No don't change what is sent for existing events, it should only be for HOME_PASSKEY_DELETE_SUCCESSFUL and HOME_PASSKEY_DELETE_FAILED

package-lock.json
liegeandlief, line 1: Is there a reason this has changed in this PR?

martin-gwar: Its basically a minor version update from 5.0.5 to 5.0.6, i ran "npm install": 

`-      "version": "5.0.5",`
`-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.5.tgz",`
`+      "version": "5.0.6",`
`+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.6.tgz",`

liegeandlief: As it is unrelated to the ticket can it be removed from this branch? When installing dependencies locally it is best to use `npm run install-all` which installs everything needed and instead of running `npm install` runs `npm ci` which installs the exact dependencies in the lockfile.

martin-gwar: Apologies mate, this was as a result of running "npm audit fix" as it was failing `npm audit`
`NPM Audit................................................................Failed`
Not `npm install`

liegeandlief: Ah okay that makes sense

liegeandlief (approved):

PR #3025, opened 2026-06-04
OLH-3702: AMC authorisation events

danacotoran:
### What changed

Add new AMC authorisation audit events as specified in the ticket [OLH-3702](https://govukverify.atlassian.net/browse/OLH-3702).

Referenced [this page](https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/6128402452/Passkeys+Audit+Events#HOME_AMC_AUTHORISATION_REQUESTED) to determine which properties should be included.

## How to review
Code review and double check the events have been set up correctly.

[OLH-3702]: https://govukverify.atlassian.net/browse/OLH-3702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/amc-callback/amc-callback-utils.ts
```diff
@@ -190,66 +203,79 @@ export async function handleJourneyOutcomeResponse(
...
+  service.send(homeActionEvent, res.locals.trace);
+
+  const homeAmcAuthorisationReceivedEvent = service.buildAuditEvent(
+    req,
+    res,
+    EventName.HOME_AMC_AUTHORISATION_RECEIVED,
+    {
+      amc_scope: journeyAction,
+      account_action_overall_success: success,
+      account_actions: [journeyAction],
```
liegeandlief: The HOME_AMC_AUTHORISATION_ERROR_RECEIVED event shouldn't have `extensions.account_actions`. See the thread at https://gds.slack.com/archives/C02F70D2LQN/p1779878710400729

liegeandlief: The AMC response may contain multiple journey actions and they should all be included here

liegeandlief: The journey outcome can contain multiple journey actions and they should all be included here

src/components/amc-callback/amc-callback-utils.ts
```diff
@@ -202,72 +203,109 @@ export async function handleJourneyOutcomeResponse(
   const service = eventService();
 
   let journeyAction: JourneyAction | undefined = undefined;
+  const accountActionsFailed =
+    actions.filter((obj) => obj.success === false).map((obj) => obj.action)
```
liegeandlief: The map is unnecessary here

src/components/amc-callback/amc-callback-utils.ts
```diff
@@ -202,72 +203,109 @@ export async function handleJourneyOutcomeResponse(
...
+      .length > 0;
 
   if (isPasskeyCreateJourney) {
     journeyAction = JourneyAction.PASSKEY_CREATE;
   }
 
+  const actionCompletedAuditEventParams: Extensions = {
+    account_action: journeyAction,
+    account_action_overall_success: success,
```
liegeandlief: In this case this is the success of the individual action so shouldn't use the top-level journey success value.

I think a good approach would be to remove it here and then prior to each `sendJourneyOutcomeEvents` call set it as follows:

- passkey creation success: true
- user logged out: false
- passkey create user aborted journey: false
- unknown error: false

src/components/amc-callback/amc-callback-utils.ts
```diff
@@ -202,72 +203,109 @@ export async function handleJourneyOutcomeResponse(
...
     res.status(HTTP_STATUS_CODES.INTERNAL_SERVER_ERROR);
     res.render("common/errors/500.njk");
   }
 }
+
+function sendJourneyOutcomeEvents(
+  req: Request,
+  res: Response,
+  service: EventServiceInterface,
+  sharedExtensionParams: Extensions,
```
liegeandlief: Would this be better named as `actionCompletedExtensionParams`?

src/components/amc-callback/amc-callback-utils.ts
```diff
@@ -202,72 +203,109 @@ export async function handleJourneyOutcomeResponse(
...
+    { ...sharedExtensionParams }
+  );
+  service.send(homeActionEvent, res.locals.trace);
+
+  const homeAmcAuthorisationReceivedEvent = service.buildAuditEvent(
+    req,
+    res,
+    EventName.HOME_AMC_AUTHORISATION_RECEIVED,
+    {
+      ...amcAuthReceivedExtensionParams,
```
liegeandlief: I don' think there's no need to spread this into a new object

src/components/amc-callback/amc-callback-utils.ts
```diff
@@ -202,72 +203,109 @@ export async function handleJourneyOutcomeResponse(
...
+  res: Response,
+  service: EventServiceInterface,
+  sharedExtensionParams: Extensions,
+  amcAuthReceivedExtensionParams: Extensions
+): void {
+  const homeActionEvent = service.buildAuditEvent(
+    req,
+    res,
+    EventName.HOME_ACTION_COMPLETED,
+    { ...sharedExtensionParams }
```
liegeandlief: I don' think there's no need to spread this into a new object

liegeandlief (changes requested):
Just a few small things to address

src/components/amc-callback/amc-callback-utils.ts
danacotoran: Good spot, I had originally created a separate object for the properties shared by both events so the spread was to merge them together. I changed my mind about that, but I forgot to update this to reflect the change in the approach.

liegeandlief (approved):

PR #3044, opened 2026-06-10
OLH-4245: Amend activity history navigation 

danacotoran:
### What changed

------ 

#### On Activity history pages 

------ 
Requirements/context: 
Previously the "Back" link on the activity history area used to behave like the browser back button, where clicking it would lead back to the previous page in activity history, or to security if the user has currently just landed on the activity history area.

After the changes introduced in this PR, the "Back" link has been hard coded to lead back to Security, in line with the new Sign in details page.

Changes: 
- remove backLinkJS flag and associated client-side script which had been introduced specifically to handle the requirement that Back link navigation through the Activity history area should work the same as the browser back button
- add query string parameters to the change password journey link: `from` must be equal to `activity-history` and `page` a valid numeric string 

------ 

#### On the **Enter password** page

-------

Requirement: 
Both "Back" and "Cancel and go back" should return to the page in activity history where the user started the "Change password" journey.

Changes summary:
- add `from` and `page` to change password redirect url

------ 

#### On the **Change password** page

-------

Requirement: 
"Cancel and go back" should return to the page in activity history where the user started the "Change password" journey. 

Changes summary:
- update form validation middleware to pass in options when a validation error occurs 
- add `from` and `page` to confirmation page redirect url 

------ 

#### On the **Change password confirmation** page

-------

Requirement: 
The button should return to the page in activity history where the user started the "Change password" journey. 

Changes summary:
- refactor template to remove duplication and move logic to backend
- Note: for some reason the code that handles rendering the confirmation pages for the add MFA journeys lives somewhere else. I didn't change this as part of this PR seeing as it's already too big.

### Testing

### Sign-offs

### Monitoring

Changes have been made to components that are part of several journeys (e.g. confirmation page) which carries some risk

## How to review
- code review
- point out any missing test coverage
- it might be helpful to double check that nothing has been broken on other journeys (the version of the "change password" journey that starts from Security/Sign in details plus other journeys where the `update-confirmation` template is used).


src/components/change-password/change-password-controller.ts
```diff
@@ -15,7 +15,7 @@ import {
   setOplSettings,
 } from "../../utils/opl.js";
 import { MetricUnit } from "@aws-lambda-powertools/metrics";
-
+import { getRenderOptions } from "../enter-password/enter-password-controller.js";
```
liegeandlief: Using the renderOptions function for a different controller seems a bit risky. If someone changes that function's behaviour it relies on them knowing it might have a knock-on effect elsewhere

integration-tests/tests/steps/@activity-history.ts
```diff
@@ -0,0 +1,46 @@
...
+    ).href;
+    await page.goto(enterPasswordUrl);
+    await page.waitForLoadState("load");
+
+    const urlObj = new URL(page.url());
+    expect(urlObj.pathname).toBe("/enter-password");
+    expect(urlObj.searchParams.get("from")).toBe("activity-history");
+    expect(urlObj.searchParams.get("page")).toBe(pageNumber);
+  }
+);
```
liegeandlief: Are you sure this function opens a new tab? I thought you had to call Playwright's `newPage` function to get get a new tab

integration-tests/tests/features/@activity-history.feature
liegeandlief, line 1: Amazingly thorough tests!

integration-tests/tests/steps/@activity-history.ts
danacotoran: You're right this doesn't actually open in a new tab. I introduced this when I was attempting to solve this with JS. I don't remember the exact issue I came into when I tried using the proper way to open a link in a new tab with playwright, but it was something to do with the fact that doing so creates a new Page object context. 

All that said, I'm not sure this scenario is needed anymore. 
As these are all normal links now, it doesn't really matter if a user chooses to open the link in a new tab or not – the information is all contained in the query string params. So it might make sense to remove this scenario.

src/components/change-password/change-password-controller.ts
danacotoran: I thought the same. But then sonar was failing on code duplication as the "enter password" and "change password" versions of this function were nearly identical. I'd be happy to keep these separate if there is a way to switch that off

liegeandlief: Could it be extracted into a shared helper function which both controllers can call?

liegeandlief (approved):

PR #3063, opened 2026-06-16
OLH-4201: Add inactive account saved user notification banner

danacotoran:
### What changed
Display a notification to the user when they log in after saving their inactive account.
 If there is an entry for the current user subject id in the user notification table, the banner should be displayed once, then the entry should be removed from the table. 

The "success" type banner had `role="alert"` and autofocus by default - this has been removed at UCD request. If the accessibility specialist advises otherwise, the accessibility attributes can be tweaked subsequently. 

### Why did it change
The oldest accounts will be up for deletion very soon. Users will be able to save their inactive accounts from deletion by simply logging in. This is part of the solution.

### Sign-offs

## How to review
Tested on dev by populating the `user_notification` table manually for a specific user, and signing in as that user. 

It was also tested locally using the provision script.

https://github.com/user-attachments/assets/3635dfc8-c0f2-40be-8ded-9c0ae6c803bd

Please code review and highlight if there is anything I missed/automated test coverage has missed


src/components/your-services/your-services-controller.ts
```diff
@@ -12,6 +22,62 @@ const contentIds: Record<string, string> = {
...
+      internalCommonSubjectId: { S: subjectId },
+    },
+  });
+
+  try {
+    const response = await dbService.deleteItem(commandParams);
+    return response;
+  } catch (err) {
+    logger.error(
+      `Failed to delete user notification for user with subject ID ${subjectId}: ${err}`
```
liegeandlief: Is the subject ID considered PII and therefore shouldn't be in our logs?

liegeandlief (commented):
One small comment but otherwise LGTM

src/components/your-services/your-services-controller.ts
danacotoran: Ohh good point! I will remove it.

liegeandlief (approved):

PR #3076, opened 2026-06-24
BAU: Fix FlexSearch Worker thread memory leak

mattvot:
### What changed

Terminate the old FlexSearch Worker thread before replacing the search index during hourly recreation.

### Why did it change

`recreateSearchIndexes()` creates a new FlexSearch `Worker` (a `node:worker_threads` thread) every hour per locale. The old Worker reference was overwritten without terminating the thread. Worker threads are not garbage collected when dereferenced - they must be explicitly terminated. This caused a steady memory leak of one orphaned thread per locale per hour.

### Related links

- https://github.com/nextapps-de/flexsearch
- FlexSearch does not expose a `destroy()` method on Worker instances despite the type definitions suggesting otherwise. The underlying `worker_threads.Worker.terminate()` must be called directly.

### Testing

- Unit test added verifying `terminate()` is called on the Worker prototype during index recreation

### Monitoring

## How to review

Single commit. The fix is 3 lines plus a type definition to properly type the internal `worker` property that FlexSearch exposes at runtime but not in its type definitions.

### DIFF
--- a/src/components/search-services/search-services-controller.ts
+++ b/src/components/search-services/search-services-controller.ts
@@ -3,6 +3,7 @@ import { getAppEnv, getClientsToShowInSearch } from "../../config.js";
 import { LOCALE } from "../../app.constants.js";
 import { MetricUnit } from "@aws-lambda-powertools/metrics";
 import { Worker, Index } from "flexsearch";
+import { Worker as ThreadWorker } from "node:worker_threads";
 import { getTranslations } from "di-account-management-rp-registry";
 import i18next, { TFunction } from "i18next";
 import { safeTranslate } from "../../utils/safeTranslate.js";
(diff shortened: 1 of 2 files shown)


src/components/search-services/search-services-controller.ts
```diff
@@ -62,7 +63,11 @@ export const getAllServices = (translate: Request["t"], locale: LOCALE) => {
     });
 };
 
-const indexes: Record<string, Index<true, false, true>> = {};
+type WorkerIndex = Index<true, false, true> & {
+  worker?: ThreadWorker;
```
liegeandlief, line 67: Why do we need to augment the `Index` type from flexsearch? Will there actually ever be a worker property on the index created by flexsearch? I wonder if these changes will actually do nothing because there will never be a `worker` property on which to call the `terminate` function

PR #3080, opened 2026-06-27
Fix TypeScript type roots and auth middleware error handling

mohsinsurani:
## Summary

This fixes two small TypeScript safety issues that show up most clearly in editor and tsserver workflows:

- make the custom type roots project-relative so VS Code and TypeScript can consistently resolve both local Express augmentations and dependency types
- narrow caught auth middleware errors before reading `message`
- pass OIDC client initialisation failures to Express error handling after recording the discovery-unavailable metric

## Context

The repo has custom Express request augmentation under `@types/express` for properties such as `req.oidc`, `req.metrics`, and `req.csrfToken`. The existing `typeRoots` entry included `/node_modules/@types`, which points at the filesystem root rather than this project. That can make local editor type resolution brittle even when the project build works.

While looking at the affected middleware, the auth error path also had an unsafe `error.message` access. This change keeps the existing metric behaviour for OIDC discovery failures, makes the catch block type-safe, and ensures the error is not silently swallowed.

## Testing

- `npm run build-ts -- --pretty false`
- `npm run test:unit -- test/unit/middleware/auth-middleware.test.ts`
- `npx prettier --check src/middleware/auth-middleware.ts test/unit/middleware/auth-middleware.test.ts tsconfig.json`


mohsinsurani:
Hi, just checking in on this one.

It’s a small TypeScript cleanup that fixes the `typeRoots` path and makes the auth middleware error handling a bit safer. I’ve added a focused test for the changed behaviour.

Happy to tweak anything if there’s a preferred direction.

PR #3095, opened 2026-07-02
OLH-4233: Playwright test tweaks

danacotoran:
### What changed

`networkidle` (the event the `the page has finished loading` step is configured to wait for) doesn't necessarily verify that the page has loaded in the sense we might expect it. It resolves when there are no network requests for at least 500 ms which might be needed on pages where Javascript or CSS is injected into the DOM dynamically (as is the case for the webchat component on the contact page) but not necessarily useful or necessary on most of our other pages. 

This removes "And the page has finished loading" from most test scenarios, relying on Playwright's built-in auto waiting capabilities instead.

Update integration tests workflow to upload test results as an artifact on Github Actions.

### Why did it change

Speed up integration test runs a little (hopefully).

The test result artifact being attached to integration test action runs (see [Artifacts produced on this run](https://github.com/govuk-one-login/di-account-management-frontend/actions/runs/28584231943)), should help diagnose the issues we are currently having where tests are passing locally yet occasionally failing in Actions.

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -97,7 +97,7 @@
       "cancelButton": "Cancel and go back"
     },
     "removePasskeyConfirmation": {
-      "title": "You’ve removed your paskey",
+      "title": "You’ve removed your passkey",
       "panelText": "You’ve removed your passkey",
       "summaryText": "Next time you sign in, you can use another passkey.",
       "summaryTextNoPasskeys": "Next time you sign in, you’ll need to enter your password and a security code."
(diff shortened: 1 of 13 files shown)


src/locales/en/translation.json
```diff
@@ -97,7 +97,7 @@
       "cancelButton": "Cancel and go back"
     },
     "removePasskeyConfirmation": {
-      "title": "You’ve removed your paskey",
```
liegeandlief, line 100: Good spot!

liegeandlief (approved):

PR #3103, opened 2026-07-08
BAU: Sign out button double underline fix

danacotoran:
### What changed

I noticed a double thickness underline is applied to the sign out button when both focus and hover are applied. Moving the focus rule after the hover rule should fix this seeing as `govuk-focused-text` removes `text-decoration`.

#### Before

https://github.com/user-attachments/assets/96970186-2cd4-4e20-a32e-5a6015c3788f

#### After

https://github.com/user-attachments/assets/e5c01416-ecbe-4368-955b-0a377fb81805


liegeandlief (approved):
Well noticed, I think I must have introduced that when making this change https://github.com/govuk-one-login/di-account-management-frontend/pull/3102

PR #3121, opened 2026-07-15
OLH-4481: bump analytics lib

liegeandlief:
Bumps the frontend analytics library to the latest version


mattvot (changes requested):
Just a blocking comment until we are given the OK to merge

mattvot (approved):

PR #3146, opened 2026-07-29
OLH-4467: Replace axios with fetch

danacotoran:
### What changed

Update http client to replace axios with native fetch
- re-implement the custom Http client wrapper using native fetch
- remove the redundant `proxy: false` property
- implement error handling with more detailed logging
- handle/log errors manually with more detail
- replace the native Axios timeout 10000 configuration by using a setTimeout block with an AbortController signal to cancel hanging requests after 10 seconds

### Why did it change
Keep project more lightweight with fewer external dependencies.
Reduce our attack surface (axios has had critical security vulnerabilities).
Improve logging when requests go wrong 

### Monitoring

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,6 @@
         "@govuk-one-login/frontend-ui": "^4.2.3",
         "@govuk-one-login/frontend-vital-signs": "0.1.4",
         "@simplewebauthn/browser": "^13.3.0",
-        "axios": "^1.18.1",
         "base64url": "3.0.1",
         "body-parser": "^2.3.0",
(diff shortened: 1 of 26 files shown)


liegeandlief (changes requested):
I deployed this branch to dev as I wanted to do some manual testing before approving it. But it seems this branch is missing the changes from https://govukverify.atlassian.net/browse/OLH-4505 which is therefore causing it to break on dev

danacotoran:
Apologies @liegeandlief, I hadn't updated it with main for a while but it should be up to date now.

liegeandlief (approved):

PR #3148, opened 2026-07-30
OLH-4463: Remove connect-dynamodb

alex9smith:
### What changed

Replace the unmaintained `connect-dynamodb` library with an express-session compatible class which wraps the create, update and delete operations. The items written to DynamoDB are the same format as `connect-dynamodb`, so this can be deployed with zero downtime.

This class adds a new feature over `connect-dynamodb` where we can now not provide a `maxAge` parameter in the cookie options and therefore get a browser-session scoped cookie which will be deleted when users close their browser tab or window. I've not used the feature yet - the cookie expiry is still passed through from the environment variable. We'll make that change in a future PR.

In the future we may want to reuse this code in other frontends. I've intentionally kept the new class with the generic operations separate from the application specific ones (which are still in `session-store.ts`) and passed all the config into the class' constructor. This'll make it easier to pull out later on.

### Why did it change

`connect-dynamodb` is unmaintained and doesn't support setting a cookie with no `maxAge` parameter.

### Related links

The [same implementation in AMC](https://github.com/govuk-one-login/account-components/blob/main/solutions/frontend/src/utils/dynamoDbSessionStore.ts).

### Monitoring

## How to review

I've deployed this to dev, tested it all behaves sensibly and looked at the items in DynamoDB before and after signing in. It'd be worth the reviewer doing the same.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,6 @@
         "axios": "^1.18.1",
         "base64url": "3.0.1",
         "body-parser": "^2.3.0",
-        "connect-dynamodb": "^3.0.5",
         "cookie-parser": "^1.4.7",
         "csrf-sync": "^4.2.1",
         "di-account-management-rp-registry": "github:govuk-one-login/di-account-management-rp-registry#main",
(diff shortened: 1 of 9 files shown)


liegeandlief (approved):
Nice one, it's really good to get rid of that dependency!

alex9smith:
I'm going to wait until Monday to merge this, just to be extra cautious

